PR #19, opened 2018-04-23
Added support for the Turcom TS-6610

different55:
Edited the Huion H610's file to add support for the Turcom TS-6610 tablet.

### DIFF
--- a/data/huion-h610-pro.tablet
+++ b/data/huion-h610-pro.tablet
@@ -4,7 +4,7 @@
 
 [Device]
 Name=Huion H610 Pro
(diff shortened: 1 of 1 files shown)


whot:
Looks good. @jigpu the coverity run seems to fail, want to fix that before merging or merge anyway?

jigpu:
The Travis issues are due to a problem on Coverity's side. Their script returns an error code if the HTTP status code after upload is anything other than 201 ("Created") but their server is returning code 200 ("OK"). I notified Synopsys of the issue a few weeks back but it doesn't look like they've fixed things yet.

Since its not an issue with the patch, I'll just merge anyway :)

PR #23, opened 2018-05-10
Add entries for graphire4 6x8 tablet

berrange:
This data was copied from the graphire4 4x5 tablet data file with the obvious changes made wrt USB ID and size.

I am unsure, however, if this is entirely correct. eg the data file says 2 buttons, but the mouse instrument has 3 buttons. Empirically the mouse left button seems to trigger same event as a stylus tap,  right button triggers same event as the bottom stylus button, and middle button triggers same event as top stylus button.  

The 2 buttons on the tablet itself though don't generate any button events under wayland. The wheel scroll on the tablet and mouse also don't generate anything.  So perhaps I'm missing some key data in this file ?

### DIFF
--- a/data/graphire4-6x8.tablet
+++ b/data/graphire4-6x8.tablet
@@ -1,0 +1,33 @@
+# Wacom
+# Graphire4 6x8
+# CTE-640
+#
+# Button Map:
+# (A=1, B=2, C=3, ...)
+#
+#               AA  BB
+#      *----------------------*
+#      |                      |
+#      |                      |
+#      |                      |
+#      |        TABLET        |
+#      |                      |
+#      |                      |
+#      |                      |
+#      *----------------------*
+#
+[Device]
+Name=Wacom Graphire4 6x8
+DeviceMatch=usb:056a:0016
+Class=Graphire
+Width=8
+Height=6
+Layout=graphire4-6x8.svg
+
+[Features]
+Stylus=true
+Buttons=2
+
+[Buttons]
+Top=A;B
+EvdevCodes=0x116;0x115
--- a/data/layouts/graphire4-6x8.svg
+++ b/data/layouts/graphire4-6x8.svg
@@ -1,0 +1,66 @@
+<?xml version="1.0" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
+   "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<!--
+   Designed after data from http://www.wacom-asia.com/download/manuals/G4_QSG_EN.pdf
+   Size and positions of controls may not be accurate
+ -->
+<svg
+   xmlns="http://www.w3.org/2000/svg"
+   version="1.1"
+   style="color:#000000;stroke:#7f7f7f;fill:none;stroke-width:.25;font-size:8"
+   id="graphire4-4x5"
+   width="208"
+   height="204">
+  <title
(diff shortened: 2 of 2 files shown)


berrange:
Oh, my mistake wrt tablet buttons was that I forgot to update udev rules. The buttons do work, but they are flipped. The left button appears to be treated as right button, and vica-verca.

Changing the .tablet file to say  Top=B;A instead of Top=A;B doesn't seem to make a difference

whot:
Use the new `EvdevCodes` field for the buttons, that should make them remap correctly, it's a list of hex codes for the `BTN_...` event that each button sends (in order A, B, C, ...)

berrange:
I pushed an update which adds use of EvdevCodes field, but that doesn't make any difference to how GNOME settings interprets the buttons - it still has them reversed.

I also noticed that the stylus buttons are reversed in GNOME settings too - it is confused between top & bottom style buttons when assigning actions.

data/graphire4-6x8.tablet
```diff
@@ -0,0 +1,33 @@
+# Wacom
+# Graphire4 6x8
+# CTE-440
```
jigpu: This needs to be updated to match the model number of your tablet (CTE-640, I believe?)

jigpu:
Your patch looks correct aside form the one minor change to the model number. The codes sent by the kernel should indeed be `0x116,0x115`, and the order should indeed be `A;B`.

The problem you're having with the pad buttons is likely a bug in GNOME. The control panel has historically had issues with Graphire/Bamboo devices because they send different X11 codes than expected. Your Graphire is programmed to have buttons A and B work as back/forward buttons, so our X driver sends buttons 8 (back) and 9 (forward). GNOME on the other hand expects "A" to send button 1 (left), "B" to send button 2 (middle), etc. The new `EvdevCodes` property was introduced to fix that, so that GNOME could see that your tablet will be sending back/forward codes. GNOME has not yet added support for the `EvdevCodes` property though, so defining it won't do much good at the moment unfortunately.

There's a workaround to have GNOME recognize the correct button sequence, but it requires special xf86-input-wacom configuration. Let me know if you're interested and I'll find it.

berrange:
I've just pushed update with fixed model name. Thanks for the detailed explanation about the buttons. I'm using Wayland so an xf86 config won't help, but no matter, I don't really need that problem with GNOME fixed in any hurry, so I'll just wait. I just wanted to make sure I understood where the problem was & whether it was known about to maintainers.

jigpu:
Ah, I see. [GNOME bug #770464](https://bugzilla.gnome.org/show_bug.cgi?id=770464) is open and suggests they start using `EvdevCodes` to fix this issue. You might chime in there to let them know you'd like to see this fixed as well.

As for your issues with the stylus buttons, I've just tested this out in Wayland with my Cintiq and see similar issues: selecting middle- or right-click in gnome-control-center does the opposite thing, and with the opposite button (e.g., set both buttons to "back" and press each of them over top of the sidebar in Nautilus: nothing happens, as expected. Next set the the "top button" to middle click and try again: a *right* click occurs when pressing the *bottom* button. Finally set the "bottom button" to right click: a *middle* click occurs when pressing the *top* button). I don't see any issues filed at GNOME about this, so I'd go ahead and report it at https://gitlab.gnome.org/GNOME/gnome-control-center. Be sure to provide us with a link to the bug so that others can find it :)

jigpu:
Request merged, and then quickly force-pushed with my Reviewed-by as c45870f (oops!)

PR #29, opened 2018-07-19
Don't double-free of the name during matching

whot:
The name is freed a few lines down as we fall through to the 'bail:' label.
The only exception is when we actually return the device, that's where we need
to free it manually.

This is just to shut up coverity, this code can only be triggered by supplying
an invalid WacomFallbackFlag.


Pinglinux:
It makes sense to me.

Reviewed-by: Ping Cheng <ping.cheng@wacom.com>

whot:
thanks!

PR #31, opened 2018-08-03
libwacom.c :: compare_matches: fix nested loop

Dico200:
The nested loop of this comparison function looked malformed, 
as it was always comparing only against the first match by incrementing and comparing to the wrong variable.

So I decided to propose this change.

### DIFF
--- a/libwacom/libwacom.c
+++ b/libwacom/libwacom.c
@@ -364,7 +364,7 @@ compare_matches(const WacomDevice *a, const WacomDevice *b)
 
 	for (match_a = ma; *match_a; match_a++) {
 		int found = 0;
-		for (match_b = mb; !found && *mb; mb++) {
+		for (match_b = mb; !found && *match_b; match_b++) {
 			if (strcmp((*match_a)->match, (*match_b)->match) == 0)
 				found = 1;
 		}


Dico200:
Just out of curiosity, does this impact anything? Surely this would've been found as certain expected behaviour would not be seen because of missing matches otherwise.

whot:
I don't think so. The only user of this function seems to be our own `dbverify` test. I grepped mutter, gnome-control-center and libinput which are the only users of libwacom that I know of and none of those use `libwacom_compare()`, let alone with the specific argument required here.

Why the `dbverify` tool didn't pick this up I don't know yet, but it's on my todo list to figure that out. If you beat me to it, I'll gladly merge any patches ;)

whot:
ok, figured it out:  this function is only called during the dbverify test that checks that the contents of the database are the same after writing it out. This means that for any device, the matches have to be the same, that's what the loop is supposed to test.

But we write out the database so that it's literally identical, including the order of the matches. The loop was supposed to be a double-nested loop, checking each A whether any of B matches. This is how it works now with your fix.

It worked before because the matches are aligned in the same order. Because we're guaranteed that the first A matches the first B, we exited that loop but increased `mb` as part of the for loop. `mb` now points to the second match in B. Comparing the second A succeeds because we point to a sublist where the exact match is the first entry again. And so on.

This would've broken if we'd write the match order in any different way. But as it was, it just happend to work.

Dico200:
Well, excuse me for not fully understanding the reasons. I briefly looking at the repository in relation to
an open source third party tablet driver that is used mainly by osu! players: https://github.com/hawku/TabletDriver

My friend, who has the newer wacom tablet whose model number is 4 digits instead of 3, I don't remember, was struggling to use the driver despite it supporting that model.

The driver referred to this repository as the source of the tablet properties.

whot:
libwacom describes properties that the kernel itself doesn't have available - things like "is it a built-in tablet" and the order/location of buttons. libwacom doesn't handle events at all, see also https://who-t.blogspot.com/2017/10/what-is-libwacom.html for some background info. As a general rule, whether libwacom knows about a tablet or not doesn't affect the basic functionality of the tablet, only fancier features like automatic screen mapping.

The model numbers (CTH-1234) don't matter at all, the only thing that matters is the USB ID and that is already 4 hex digits anyway in the form of `bustype:vendor-id:product-id`, the two IDs are 16 bit numbers and must be written as 4-digit hex in the libwacom matches.

I can't find any reference to libwacom in that repo you linked though (at least not in the code or the wiki).

Dico200:
Oh, it was referring to `input-wacom` repo and then I must have gotten sidetracked.
[File](https://github.com/hawku/TabletDriver/blob/master/TabletDriverService/config/wacom.cfg)

Issue #36, opened 2018-09-07
Add Lenovo ThinkPad X360 Yoga


jigpu:
Sysinfo at https://github.com/linuxwacom/wacom-hid-descriptors/issues/38

PR #40, opened 2018-09-07
data: Add ISDv4 5150 (Lenovo ThinkPad X380 Yoga)

jigpu:
Ref: https://github.com/linuxwacom/libwacom/issues/36

### DIFF
--- a/data/isdv4-5150.tablet
+++ b/data/isdv4-5150.tablet
@@ -1,0 +1,14 @@
+# this is for the Wacom pen + touchscreen as found in some versions of the Lenovo ThinkPad X380 Yoga
+
+[Device]
+Name=Wacom ISDv4 5150
+DeviceMatch=usb:056a:5150
+Class=ISDV4
+Width=12
+Height=7
+IntegratedIn=Display;System
+
+[Features]
+Stylus=true
+Touch=true
+Buttons=0


maksbotan:
Hi Jason,

Unfortunately, it does not work. I placed that file in `/usr/share/libwacom`, rebooted and still got nothing in `libwacom-list-local-devices`. And it still generates no events and does not work in Gimp.

My device is ThinkPad Pen Pro included with my laptop, I suppose it's charged, since it's constantly inside the laptop and so connected to the charger.

What additional info can I provide?

I noticed that it also did not work in Windows (at least it did not generate mouse clicks, I did not try drawing). Maybe there's something I'm missing?

whot:
@maksbotan does it show up in `sudo evemu record` and if so, does it send events? If not, then it's a kernel issue or possibly a hardware issue because if it doesn't work in windows either, then that's the most likely cause.

libwacom merely *describes* the device, it has no effect on whether it works or not. The only benefit you'll get from the libwacom is that it'll show up in the GNOME control center and screen mapping etc. suddenly automagically works.

maksbotan:
evemu recognizes the pen:

```
# EVEMU 1.3
# Kernel: 4.18.5-gentoo
# DMI: dmi:bvnLENOVO:bvrR0SET29W(1.13):bd04/16/2018:svnLENOVO:pn20LH000MUS:pvrThinkPadX380Yoga:rvnLENOVO:rn20LH000MUS:rvrSDK0J40697WIN:cvnLENOVO:ct31:cvrNone:
# Input device name: "Wacom Pen and multitouch sensor Pen"
# Input device ID: bus 0x03 vendor 0x56a product 0x5150 version 0x111
# Size in mm: 293x165
```

However, no events are shown. Maybe I'll try checking it in Windows once more.

jigpu:
I've just pushed a new version of this patch which should fix the issue you had. My original definition specified the sensor was connected over I2C, but your device is actually connected over USB. Please see if the [updated file](https://raw.githubusercontent.com/linuxwacom/libwacom/49c6138776081126197528c86634f7a34509cf66/data/isdv4-5150.tablet) causes the device to show up in `libwacom-list-local-devices`

jigpu:
Okay, thanks :) If you do see the pen working under Windows, open a new issue on the [input-wacom tracker](https://github.com/linuxwacom/input-wacom/issues) and we'll try to work out what's going on.

maksbotan:
Pen does not seem to work on windows as well :/
I fear this is not the right place, but I don't know better. Can you give an advice how can I debug this?

My device is Thinkpad Yoga x380 with pen enclosed in the laptop itself. The pen is called "thinkpad pen pro" and has two buttons. It also has two contacts, for charging I presume.

Oh, and I wanted to ask is there a way to enable multitouch gestures with fingers? For example, pinch-to-zoom in chrome? It works on windows.

jigpu:
Unfortunately I'm not too sure about Windows troubleshooting... There's a chance that its internal battery/supercap died or that there's some other defect with the pen. If you know anyone else with a Thinkpad pen, you could see if theirs works since it should be compatible. Failing that, I'd get in touch with Dell and go through their troubleshooting steps. They should be able to either get the pen working under Windows (which would, hopefully, also cause it to work under Linux) or help you get a replacement pen if they think its broken.

As for gestures, you can try using `xsetwacom set <touch_id> gesture off` with the name or ID of the touchscreen device. This disables in-driver gesture support which will allow applications like Chrome to do their own gesture handling. The in-driver scroll/zoom gestures should work in basically any app, but they're not nearly as slick as what some applications are capable of.

whot:
I'm going to merge this because it shows up in `libwacom-list-local-devices` now, libwacom won't have an effect on whether the pen works anyway. Thanks

PR #42, opened 2018-09-11
data: Add ISDv4 502A (Toshiba Portege X20W-D Convertible and Toshiba TruPen)

tom0105:
Add support for the Toshiba Portege X20W-D Convertible and Toshiba TruPen


whot:
thanks!

PR #44, opened 2018-09-16
Add issue templates for github bug reports

whot:
The hope here is that it cuts down on reporters filing issues in libwacom that then need to be moved to input-wacom, xf86-input-wacom or libinput. Or /dev/null ;)

Happy to modify the templates as needed


jigpu:
Adding templates to the other projects (to deal with the same issue of bugs being filed in the wrong project) has been on my to-do list for a while. Totally in favor of this.

whot:
thanks! of course feel free to steal the templates or whatever is useful of them for the other projects

PR #48, opened 2018-10-05
Add support for EDID information for integrated tablets

### LINKED ISSUE #47: Add Monitor EDID to the data files
g-s-d/gnome-control-center/mutter have heuristics to pick the right monitor for the various tablet devices. For external devices that's guesswork anyway but for internal devices we could make their life a lot simpler by just providing some monitor information that we know matches the device. This should make it possible to reliably identify Cintiqs to their respective monitor.

cc @bentiss because he came up with this  idea. 

cc @garnacho, @jigpu

whot:
Provide just enough data about the monitor we're integrated in to take the
guesswork out of the game for anyone who needs to map the tablet to the right
monitor.

Fixes #47 

cc @garnacho, @bentiss

### DIFF
--- a/data/wacom.example
+++ b/data/wacom.example
@@ -97,6 +97,42 @@ IntegratedIn=Display;System
 # Filename of the SVG representing this tablet. Only needed for tablets with
 # pad buttons.
 Layout=bamboo-16fg-m-pt.svg
+
+# EDID information if IntegratedIn is Display or System.
+# These are the 8 bytes in the EDID 1.4 struct comprising manufacturer ID,
+# product code and serial number. For reference, edid-decode output:
+# Extracted contents:
+# header:          00 ff ff ff ff ff ff 00
+# serial number:   10 ac 12 ab 4c 56 4a 30 16 16...
+#                  ^^^^^ ^^^^^ ^^^^^^^^^^^
+#                   Man   Prod   Serial
+#
+# The ProductName is separate in the EDID Other Monitor Descriptors as code
+# 0xFC.
+#
+# A match group is defined as string in the format: <Man>:<Prod>:<Name>:<Serial>.
+#
+# Manufacturer ID: a 3-byte string, e.g. DEL. These are three 5-bit chars from
+#                  the 15 least-significant bits of the first two bytes in the
+#                  EDID struct where 'A' == 1, 'B' == 2, etc. See the EDID
+#                  documentation. Example:
+#                  DEL == (4 << 10) | (5 << 5) | 12 == 0x10ac
+# Product code: a 4-character (16-bit) hex string in little endian order,
+#                 e.g. 12ab
+# Serial number: a 8-character (32-bit) hex string in little endian order,
+#                e.g. 4c564a30
+# Name: a string length 13 or less (shell globs supported)
+#
+# Any or all fields can be omitted using a literal '*'.
+#
+# The EDIDMatch contains one or more match groups, separated by
+# semicolons, e.g.
+# EDIDMatch=DEL:12ab:*:*;LEN:*:*:Lenovo Some Monitor;*:1234:01020304:*foo*:
+#
+# EDIDMatches are not exclusive, multiple tablets may have the same
+# EDIDMatch entries.
+EDIDMatch=DEL:12ab:*:*
+
 # Styli
 #
 # This is a list of stylus IDs supported by the tablet. Non-Wacom devices
--- a/libwacom/libwacom-database.c
+++ b/libwacom/libwacom-database.c
@@ -34,6 +34,11 @@
 #include <linux/input-event-codes.h>
 
 #include <assert.h>
+#ifdef __FreeBSD__
+#include <sys/endian.h>
+#else
+#include <endian.h>
+#endif
 #include <glib.h>
 #include <dirent.h>
 #include <string.h>
@@ -46,6 +51,7 @@
 #define FEATURES_GROUP "Features"
 #define DEVICE_GROUP "Device"
 #define BUTTONS_GROUP "Buttons"
+#define EDID_GROUP "EDID"
 
 static WacomClass
 libwacom_class_string_to_enum(const char *class)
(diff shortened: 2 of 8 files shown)


whot:
test suite failure is because I'm using a glib function that's only been in glib for a year, I'll fix this next week.

garnacho:
IMHO this would be great to have, it offloads the grittiest heuristic from gnome-settings-daemon and replaces it by something clearly spelled out. 

The patch and extension to .tablet desktop file format also make sense to me. I might suggest using `g_ascii_strtoull()` if we want to avoid the dependency on recent glib.

jigpu:
This is an interesting idea, though I'm not sure if we can reliably say that a given VID:PID is associated with a given EDID (or vice-versa). Definitely back in the days when EMR sensors were used, we would see the same PID used by multiple OEMs. Newer AES sensors seem to be specific to devices, and possibly even the panel. As long as OEMs don't use multiple different panels with different EDIDs but same the same PID, this might work. I'll have to ask around...

whot:
Force-pushed a few changes, main differences:
- use `g_ascii_strtoull()` to avoid newer dependencies
- write a parsing function so we can have tests for it and ideally not fail on the hex parsing code. because oh how I love writing parsers in C....

jigpu:
I checked around and it looks PID <-> EDID matching is *almost* a one-to-many relation with our display tablets. Quite a few recent Cintiqs have two possible EDIDs, so we'd need to be sure the database can handle that. Conversely, the vast majority of EDIDs can be mapped uniquely back to a PID, but there's at least two instances where a single EDID was used for more than one PID. I'm less sure about the situation with OEMs, but treating this as many-to-many is a safe bet.

data/wacom.example
```diff
@@ -158,3 +158,26 @@ Touchstrip2=J
 # We assume the same number of modes for each of the touchstrips
 # if there is more than one
 StripsNumModes=4
+
+# EDID information if IntegratedIn is Display or System.
+# These are the 8 bytes in the EDID 1.4 struct comprising manufacturer ID,
+# product code and serial number. For reference, edid-decode output:
+# Extracted contents:
+# header:          00 ff ff ff ff ff ff 00
+# serial number:   10 ac 79 a0 4c 56 4a 30 16 16
```
jigpu: Would be nice if the data here matched up with the example values below

whot:
force-pushed again, changes are: new format is `EDIDMatch=DEL:*:*:somename` or some combination thereof. This allows for multiple matches per device.

The matching itself is now done inside libwacom, so the caller calls `libwacom_edid_match_manufacturer_id(device, "DEL")` instead of fetching the data and then doing a comparison on it.

data/wacom.example
```diff
@@ -49,6 +49,39 @@ Height=6
...
+# 0xFC.
+#
+# A match group is defined as string in the format: <MID>:<PC>:<NAME>:<SERIAL>.
+#
+#  Manufacturer ID: a 3-byte string, e.g. DEL. These are three 5-bit chars from
+#                   the 15 least-significant bits of the first two bytes in the
+#                   EDID struct where 'A' == 1, 'B' == 2, etc. See the EDID
+#                   documentation. Example:
+#                   DEL == (4 << 10) | (5 << 5) | 12 == 0x10ac
+#  Product code: a 4-character hex string in big endian order, e.g. 12ab
```
jigpu: The EDID stores the product code in little-endian order, so I believe "12 ab" from the example hexdump would become "ab12" when converted to big-endian.

data/wacom.example
```diff
@@ -49,6 +49,39 @@ Height=6
...
+#  Serial number: a 8-character hex string in big endian order, e.g. 4c564a30
```
jigpu: Ditto here: s/4c564a30/304a564c/

libwacom/libwacom-database.c
```diff
@@ -592,6 +655,29 @@ libwacom_parse_tablet_keyfile(const char *datadir, const char *filename)
...
+		guint i;
+
+		array = g_array_new (FALSE, FALSE, sizeof(WacomEDID));
+		device->num_edid = 0;
+
+		for (i = 0; string_list[i]; i++) {
+			WacomEDID match;
+
+			if (!edid_match_from_string(string_list[i], &match)) {
+				DBG("failed to match '%s' for EDID. Skipping.\n", string_list[i]);
```
jigpu: I'd suggest using "failed to parse" instead to make it clear that this is an issue with the datafile, not the actual matching process.

libwacom/libwacom.c
```diff
@@ -350,6 +351,8 @@ libwacom_copy(const WacomDevice *device)
 	d->num_buttons = device->num_buttons;
 	d->buttons = g_memdup (device->buttons, sizeof(WacomButtonFlags) * device->num_buttons);
 	d->button_codes = g_memdup (device->button_codes, sizeof(int) * device->num_buttons);
+	memcpy(&d->edid, &device->edid, device->num_edid * sizeof (*device->edid));
```
jigpu: Is `d->edid` being assigned/allocated somewhere? It looks like its value hasn't been set after the call to `g_new0`. Segfault waiting to happen?

libwacom/libwacom.h
```diff
@@ -636,6 +637,33 @@ uint32_t libwacom_match_get_product_id(const WacomMatch *match);
...
+ */
+bool libwacom_edid_match_serial_number(WacomDevice *device,
+				       uint32_t serial_number);
+
+/**
+ * Return true if the given EDID name matches those defined for this device.
+ */
+bool libwacom_edid_match_product_name(WacomDevice *device,
+				      const char *product_name);
+
```
jigpu: I'm not sure I like the idea of matching each part of the EDID in isolation. It seems like you'd potentially wind up with false positives. E.g. a device with `EdidMatch=ABC:1234:*:*;XYZ:9876:*:*` could be matched by monitors with EDID `ABC:9876:*:*` or `XYZ:1234:*:*`.

data/wacom.example
whot: whoops, I'm gonna switch this to LE instead, no point in using BE here if EDID itself is LE

data/wacom.example
whot: see above, switching to LE is the more sensible option here.

libwacom/libwacom.h
whot: I've changed this to a custom `WacomEDID` struct that is set by the caller and then compared as a whole in the latest push

whot:
New version, rebased and re-done. The new approach is that the caller builds a `WacomEDID` struct with the information it has and then tries to match it with a device. Globs are allowed in the `.tablet` files but not in the caller for obvious reasons, but where the caller omits a field and the `.tablet` file has a glob, we match successfully. But at least one field must be set, you can't match against all tablets.

whot:
Closing this one - no apparent need for this in 2 years and I've long since paged everything out.

Issue #68, opened 2019-02-07
Cintiq Pro 2 24/ExpressKey Remote naming inconsistency


tb2097:
This is a similar issue to the one found in #64 with the values from xsetwacom not matching those provided by libwacom; I have implemented a hack to get around it for now: [[tb2097/wacom-gui/issues/26]]

Fixes required:
data/ek-remote.tablet:
15    - Name=Wacom ExpressKey Remote
15    + Name=Wacom Express Key Remote

data/cintiq-pro-24-p.tablet:
30    - Name=Wacom Cintiq Pro 24 P
30    + Name=Wacom Cintiq Pro 24

data/cintiq-pro-24-pt.tablet:
30    - Name=Wacom Cintiq Pro 24 PT
30    + Name=Wacom Cintiq Pro 24

jigpu:
The name given in ek-remote.tablet is the more correct of the two (the marketing name has no space) so I would leave it unmodified. We could potentially change the name in the kernel to match but doing so would cause user scripts to break. I think this one is best left alone.

As for the Cintiq Pro 24, removing the "P" and "PT" is reasonable enough. As mentioned in #67 those are mostly hints to us rather than actually part of the marketing name.

tb2097:
Your quick response is much appreciated.  I can leave the "hack" in for ek-remote as it's just adding the space into the name. Hopefully the *.tablet files having the p/pt in them is enough to help differentiate between them.
As well, I'm only with more people becoming aware of my tool and testing it vs having bash scripts since I've effectively aped the UI from Windows as close as I can and it supports auto-loading of configs etc, which is what most people who write them out by hand desire.  Also, I get testers for tablets I don't have access to :)

jigpu:
That's a pretty slick looking control panel :+1: I'll definitely have to give a try some time when I'm in front of one of my non-GNOME systems.
If you'd like to submit a pull request similar to #67 for the "P" and "PT" changes, you shouldn't have much trouble getting them merged.

whot:
closing, this as fixed with the PR referenced above

PR #69, opened 2019-02-08
Cintiq Pro 24 naming consistency with xsetwacom

tb2097:
Fix for [[#68]]; ExpressKeys/Express Keys issue can be skipped due to possible legacy support issues.


whot:
Merged with a modified commit message :)

PR #72, opened 2019-02-19
Add Pro Pen Slim definition

jigpu:
Wacom has released a new "Pro Pen Slim" which is compatible with the
current generation of tablets. Add its definition and update the tablet
files where necessary.


whot:
thanks!

PR #87, opened 2019-05-02
Add a tool to list the available styli

whot:
This sits on top of #86.


jigpu:
Haven't looked too closely, but seems fine enough for an Ack

whot:
FTR it's not installed anyway, so this isn't user-visible API

PR #91, opened 2019-05-21
Add tablet config for ELAN 2072 touchscreen (HP Spectre x360 13-40xx)

jkearneyma:


whot:
fwiw, the test suite basically runs `make check`, if you run that yourself you can catch most of the errors  (and it's easier to debug too)

Issue #89, opened 2019-05-15
Microsoft Surface IPTS device integration


qzed:
Hi all, I just wanted to get a discussion started about including the surface specific patches (e.g. from https://github.com/geoffholden/libwacom and https://github.com/qzed/libwacom-surface), originally created by @geoffholden.

The devices in question are Intel Precise Touch and Stylus (IPTS) devices connected via the Intel Management Engine (MEI) bus, so to integrate the devices this bus type would need to be supported ([00_mei-bus.patch](https://github.com/qzed/libwacom-surface/blob/176f72de68c6311624d9c743a2e784efba643ce5/00_mei-bus.patch)).

On the device-data side, all devices are currently listed in the same file ([01_surface-tablet-data.patch](https://github.com/qzed/libwacom-surface/blob/176f72de68c6311624d9c743a2e784efba643ce5/01_surface-tablet-data.patch)). Should we change that to be per model? Also I think it would be good if someone could look over the file and check if it's up to convention (e.g. I've added the `Styli` key and I'm not sure if that's a good idea; stylus is single-button + eraser, the eraser is also a button, but that's a separate bluetooth device).

It might also be a good idea to re-check the device IDs and map them to the specific models. Here's what I know so far:

| ID | Model | Source | Note |
|-------------|------------------------|----------------------------------------------------------------------------------------------------------------------|-------|
| `1B96:005e` | Surface Book 1 | [comment](https://github.com/jakeday/linux-surface/issues/513#issuecomment-498009154) | |
| `045E:0020` | Surface Book 2 (15") | [comment](https://github.com/jakeday/linux-surface/issues/513#issuecomment-497983039) |
| `045E:0021` | Surface Book 2 (13.5") | my own device | | |
| `1B96:006A` | Surface Pro 4 | [comment](https://github.com/jakeday/linux-surface/issues/513#issuecomment-498028173) | |
| `1B96:001F` | Surface Pro 5 (2017) |  | prev. source seems to be wrong |
| `045E:001F` | Surface Pro 6 | [comment](https://github.com/jakeday/linux-surface/issues/513#issuecomment-498036450) | |
| `04F3:261A` | Surface Go | [issue](https://github.com/linuxwacom/libwacom/issues/75#issue-421842624) | hid-over-i2c, not IPTS? |

[...]

PR #92, opened 2019-06-01
Add support for Intel Management Engine bus

qzed:
Add support for devices connected via the Intel Management Engine (MEI).
This is required to support IPTS based devices, such as (among others)
the Microsoft Surface Books, Surface Pro 5 and 6, and Surface Laptops.

As per the previous discussion in #89.

### DIFF
--- a/libwacom/libwacom-database.c
+++ b/libwacom/libwacom-database.c
@@ -112,6 +112,8 @@ bus_from_str (const char *str)
 		return WBUSTYPE_BLUETOOTH;
 	if (strcmp (str, "i2c") == 0)
 		return WBUSTYPE_I2C;
+	if (strcmp (str, "mei") == 0)
+		return WBUSTYPE_MEI;
 	return WBUSTYPE_UNKNOWN;
 }
 
@@ -130,6 +132,8 @@ bus_to_str (WacomBusType bus)
 		return "bluetooth";
 	case WBUSTYPE_I2C:
 		return "i2c";
+	case WBUSTYPE_MEI:
+		return "mei";
 	}
 	g_assert_not_reached ();
 }
--- a/libwacom/libwacom.c
+++ b/libwacom/libwacom.c
@@ -144,6 +144,10 @@ get_bus_vid_pid (GUdevDevice  *device,
 		*bus = WBUSTYPE_I2C;
 		retval = TRUE;
 		break;
+	case 68:
+		*bus = WBUSTYPE_MEI;
+		retval = TRUE;
+		break;
 	}
 
 out:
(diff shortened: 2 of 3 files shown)


whot:
this is upstream in the linux kernel already? Just confirming that we don't have to wait for that.

Also, what's the `udevadm info` output for a device like this? How does this show up in the system?

qzed:
@whot Unfortunately the IPTS stuff is not upstream, and as far as I can see likely never will without official support from Intel or some serious overhaul from a dedicated individual. I hoped you guys would be okay with that, but I can understand if you're not.

[Here](https://github.com/linuxwacom/wacom-hid-descriptors/blob/2b68fcb3fa38950584de089e67a7924c2b99a1ae/Microsoft%20Surface%20Book%202/sysinfo.zPqDidaoEx/udevadm.txt#L781-L802) and [here](https://github.com/linuxwacom/wacom-hid-descriptors/blob/2b68fcb3fa38950584de089e67a7924c2b99a1ae/Microsoft%20Surface%20Book%202/sysinfo.zPqDidaoEx/udevadm_0044:045E:0021.0001.txt#L69-L121) is the udevadm info for the tablet device, let me know if you need more info.

whot:
cc @bentiss for comments. As a general rule, I don't like pulling code that relies on out-of-tree patches because once we have it, we have to make sure we keep up with those bits as well. And possibly worry about conflicting implementations being merged at some later point and having to hack around those.

You'll need to at least get the bustype added to `linux/input.h` so no-one else can use the same number and cause conflicts later.

qzed:
@whot Right, I completely get your reservations. I somehow was under the wrong impression that the MEI bus ID was already upstream. Not sure where I got that, sorry. I can see how this would lead to conflicts. 

Assuming a fixed bus ID, I don't see any potential for any complications in the future. The dependencies of this PR are only on the bus ID and HID IDs of the individual devices, the latter being controlled by hardware.

With regards to introducing a new bus ID: I can see how there might be some push-back on introducing a new bus ID just for some out-of-tree patches. Assuming a full IPTS patch were to be upstreamed in the future, how likely would it be to get a new bus ID for it as opposed to use something like BUS_VIRTUAL or BUS_HOST? The point I want to make: Would it make more sense for us to change the bus ID in the IPTS patches to BUS_VIRTUAL or BUS_HOST and use that (until there is an upstream authority saying otherwise)?

whot:
> Assuming a fixed bus ID, I don't see any potential for any complications in the future. 

yeah, I agree with this. we support multiple matches, so as long as the prefix (i.e. bus id) differs, we can add other matches if the same devices is supported some other way upstream. The only source of conflict would be re-using bus ID and the HID IDs but that's unlikely.

Not convinced by `BUS_VIRTUAL`/`HOST` tbh, too much generalization in those.

qzed:
Alright, if there are no more comments I'll try to reserve a bus ID for MEI upstream.

qzed:
I'm going to close this and #93 for now. The IPTS driver is far from being upstreamed and is completely unusable with kernels 5.3 and 5.4 due to some changes in i915 related to GuC submission, on which it relies. Will re-submit if we've got something upstream.

PR #95, opened 2019-06-06
Tablet config for ELAN 2627

daenney:
This model is found in the HP Envy x360 ag0xxxx series.

I have a few questions though:
* Width and height, couldn't really find them specified by HP so I just used a ruler and rounded up to the closest amount of inches. Does that make sense or should I round down to 11 and 6 instead?
* The device class is giving me trouble. `PenDisplay` seems the most appropriate but if I use that the Pen doesn't work at all. Switching to Intuos4 makes the pen work but the tracking is bonkers slow, the pointer lags behind the pen a great deal, almost a second and its location is off. I also tried `Class=ISDV4` which seems to get me the same result as `Class=Intuos4`. Which one should I pick?
* Touching the screen (with my finger) seems to result in multiple events. So for example when I touch the header bar of an app it goes full-screen b/c Gnome apparently believes I double-clicked it. I'm not sure if this has anything to do with libwacom or if I should raise this with the libinput folks

When I run `libwacom-list-local-devices` I get 2, contrary to what I would expect:
```
[Device]
Name=ELAN 2627
DeviceMatch=i2c:04f3:2627;
Class=PenDisplay
Width=12
Height=7
IntegratedIn=Display;System;
Styli=0xfffff;0xffffe;

[Features]
Reversible=false
Stylus=true
Ring=false
Ring2=false
Touch=true
TouchSwitch=false
StatusLEDs=
NumStrips=0
Buttons=0
---------------------------------------------------------------
[Device]
Name=Wacom Serial Tablet WACf004
DeviceMatch=serial:0000:0000;
Class=ISDV4
Width=0
Height=0
IntegratedIn=Display;System;
Styli=0xfffff;0xffffe;

[Features]
Reversible=false
Stylus=true
Ring=false
Ring2=false
Touch=false
TouchSwitch=false
StatusLEDs=
NumStrips=0
Buttons=0
---------------------------------------------------------------
```
Is the `Wacom Serial Tablet WACf004` device supposed to be there?

Despite adding this config devices didn't show up properly for X11, I had to add this snippet:

```
Section "InputClass"
    Identifier "Elan driver override"
    MatchUSBID "04f3:*"
    MatchDevicePath "/dev/input/event*"
    MatchIsTablet "true"
    Driver "wacom"
EndSection
```

I get this from `xinput`:
```
⎡ Virtual core pointer                          id=2    [master pointer  (3)]
⎜   ↳ Virtual core XTEST pointer                id=4    [slave  pointer  (2)]
[...]

PR #96, opened 2019-06-18
tools: print the device node for list-local-devices

whot:
Makes it easier to identify where an entry comes from.

Related #95

### DIFF
--- a/tools/list-local-devices.c
+++ b/tools/list-local-devices.c
@@ -96,9 +96,12 @@ int main(int argc, char **argv)
 		char fname[PATH_MAX];
 
 		snprintf(fname, sizeof(fname), "/dev/input/%s", namelist[i]->d_name);
+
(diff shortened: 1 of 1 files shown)


daenney:
I applied this and tried it. I find the output a tiny bit confusing b/c it's not obvious, to me, whether the line that prints the device node belongs to the entry below or above it. Something like a newline at the end of a device would help make that more obvious.

jigpu:
Agreed with @daenney. Could print the filename on a commented line to be a bit more obvious (e.g. `"--------------------------------------------------\n# %s\n"`) Otherwise LGTM and would be helpful when going through sysinfo dumps.

whot:
Updated, new output is now:
```
# Device node: /dev/input/event22
[Device]
Name=Wacom Intuos3 4x6
ModelName=PTZ-431W
...
---------------------------------------------------------------
```
i.e. the device node is printed as comment at the top, rest is left as-is

PR #101, opened 2019-06-23
Ignore serial touchpads

### LINKED ISSUE #100: libwacom-list-local-devices lists serial tablets for touchpads
See #95, issue is that a serial tablet appears in the output even though there is none. This happens wherever there is a serial touchpad connected.

Cause is this hunk of code:
```
         retval = get_bus_vid_pid (device, bus, vendor_id, product_id, error);
         if (retval)
                  goto out;
  
          bus_str = get_bus (device);
          *bus = bus_from_str (bus_str);
  
          if (*bus == WBUSTYPE_SERIAL) {
                  /* The serial bus uses 0:0 as the vid/pid */
                  *vendor_id = 0;
                  *product_id = 0; 
                  retval = TRUE;
          }
```

IOW anything that's a serial device will be accepted and reset to vid/pid 0/0. That then matches the ISDV4 tablets in our database.

whot:
We don't have serial touchpad devices that we care about, but the current code ends up labeling any serial touchpad as supported device with vid/pid 0 - which in turn maps to our serial ISDV4 devices.

Fixes #100 

cc @daenney

### DIFF
--- a/libwacom/libwacom.c
+++ b/libwacom/libwacom.c
@@ -56,10 +56,22 @@ libwacom_get_device(const WacomDeviceDatabase *db, const char *match)
 }
 
 static gboolean
+is_tablet (GUdevDevice *device)
+{
+	return g_udev_device_get_property_as_boolean (device, "ID_INPUT_TABLET");
+}
+
+static gboolean
+is_touchpad (GUdevDevice *device)
+{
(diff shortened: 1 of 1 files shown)


daenney:
Neat, thanks!

PR #104, opened 2019-06-24
Add support for the meson build system

whot:
This adds the meson build system in addition to the automake bits we already have. Meson allows for a much easier to understand build system description, it's faster and many of the upstream projects we care about are already using it (libinput and GNOME in large parts).

Long-term I'd like to remove autotools but I understand that this may be off a few releases at least.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,20 @@ script:
 
 jobs:
   include:
+    - stage: test
+      install:
+        - sudo apt-get install -y ninja-build python3-pip python3-setuptools valgrind
+        - sudo pip3 install meson
+      script:
+        - meson builddir
+        - ninja -C builddir test || (cat builddir/meson-logs/testlog.txt && false)
+        - ninja -C builddir dist
+        - meson configure builddir -Db_sanitize=address,undefined
+        - ninja -C builddir test || (cat builddir/meson-logs/testlog.txt && false)
+        - meson configure builddir -Db_sanitize=none
+        - pushd builddir > /dev/null
+        - meson test --setup=valgrind --suite=valgrind || (cat meson-logs/testlog-valgrind.txt && false)
+        - popd > /dev/null
     - stage: coverity
       compiler: 'gcc'
       env:
--- a/run-full-test.sh
+++ b/run-full-test.sh
@@ -1,0 +1,52 @@
+#!/bin/bash -e
+
+date=`date +"%Y-%m-%d-%H.%M.%S"`
+builddir="build.$date"
+
+echo "####################################### running test suite"
+meson $builddir
+ninja -C $builddir test
+
+echo "####################################### running valgrind"
+pushd $builddir > /dev/null
+meson test --setup=valgrind --suite=valgrind
+popd > /dev/null
+
+echo "####################################### running ubsan"
+meson configure $builddir  -Db_sanitize=undefined
+ninja -C $builddir test
+
+echo "####################################### running asan"
+meson configure $builddir  -Db_sanitize=address
+ninja -C $builddir test
+
+echo "####################################### running scan-build"
+meson configure $builddir  -Db_sanitize=none
+ninja -C $builddir scan-build
+
+echo "####################################### running make distcheck"
+
+# we can't configure with a custom sourcetree if we have run configure
+# previously. So let's only do this where we have a clean tree that we can
+# clone.
+if git diff --exit-code -s; then
+    mkdir -p "$builddir/autotools/build"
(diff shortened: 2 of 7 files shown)


whot:
travis is still on ubuntu 16.04 which makes meson mostly impossible with our current CI script, some extra bits needed here...

whot:
I think this is ready now. Note that it sits on top of #122, only the last three commits are the meson ones.

run-full-test.sh
```diff
@@ -0,0 +1,52 @@
...
+meson configure $builddir  -Db_sanitize=undefined
+ninja -C $builddir test
+
+echo "####################################### running asan"
+meson configure $builddir  -Db_sanitize=address
+ninja -C $builddir test
+
+echo "####################################### running scan-build"
+meson configure $builddir  -Db_sanitize=none
+ninja -C $builddir test
```
jigpu: I don't understand how this is running scan-build, but I could just be missing the obvious...

.travis.yml
```diff
@@ -32,6 +32,20 @@ script:
 
 jobs:
   include:
+    - stage: test
+      install:
+        - sudo apt-get install -y ninja-build python3-pip python3-setuptools valgrind
+        - sudo pip3 install meson
+      script:
```
jigpu: I'm confused. The "test" suite normally runs the default "script" block, so adding it here makes me wonder if the autotools path is actually going to be tested now. For that matter, why don't we just have Travis run the shiny new `run-full-test.sh` script that was just introduced?

run-full-test.sh
whot: whoops. that should read `ninja ... scan-build`. Fixed, will push in a second

.travis.yml
whot: Our travis script is a bit confusing, tbh. We're using the `compiler` command which gives us a matrix of runs - clang and gcc. Those jobs run the default scripts, so we get two jobs.

Then, *in addition* we have the `jobs:` directive which are extra jobs not covered by the matrix. That's where meson now fits in. So in total we have four jobs now, the default ones run with gcc and clang, the meson job, and the coverity job (filtered to cron only). This means that meson as it is here only tests gcc, but ... well, meh? :)

Honestly, I think dropping the `compiler` directive and just duplicating the jobs with different compilers would make the travis script a lot more sensible to the occasional reader. I'm also tempted to switch this all to circleci for no other reason than that @bentiss put some effort in for libratbag and we can copy/paste most of that without having to remember yet another syntax. Unless you're having to deal with Travis elsewhere - for me it's the only project that uses it.

> why don't we just have Travis run the shiny new run-full-test.sh script that was just introduced?

I don't like running scripts in CI unless they're generic re-usable setup scripts. The commands to actually build and test should be explicit. It provides better debugging and output. And, in a rare case where duplication can be useful, having the test script separate to the CI script means we may just pick up the odd bug here or there which we wouldn't if we run exactly the same setup everywhere.

whot:
Merging this because IMO it's useful and we can fix the travis things above as follow-ups.

.travis.yml
jigpu: Ah; it makes more sense now that I see the run with this config in Travis. I didn't have time yesterday to push it up to my branch to understand what was going on.

> I'm also tempted to switch this all to circleci for no other reason than that @bentiss put some effort in for libratbag and we can copy/paste most of that without having to remember yet another syntax. Unless you're having to deal with Travis elsewhere - for me it's the only project that uses it.

FWIW, we also use Travis for xf86-input-wacom and input-wacom. It'd be nice for all the projects to share common infrastructure, but libwacom *is* primarily under your control...

> I don't like running scripts in CI unless they're generic re-usable setup scripts. The commands to actually build and test should be explicit. It provides better debugging and output. And, in a rare case where duplication can be useful, having the test script separate to the CI script means we may just pick up the odd bug here or there which we wouldn't if we run exactly the same setup everywhere.

I'm not terribly convinced by these arguments, but whatever :D I'd much rather be confident that CI isn't skipping any tests which I forgot to add in both places (or vice-versa) and can add a `set -ex` to the top of the script to get similar output/error handling.

skomra:
Just a note that Ubuntu 18.04.2 also doesn't support this version of meson:

`meson.build:1:0: ERROR: Meson version is 0.45.1 but project requires >= 0.47.0`

whot:
meson doesn't have any dependencies outside the python standard library, so it can be run directly from the build directory or installed via pip (like in the travis script that runs on 16.04). So until we drop autotools which is unlikely to happen for a while I think we're fine with a meson version not in one of the distros.

PR #107, opened 2019-06-25
Document the signed-ness mismatch between vendor ids

whot:
The vendor and product ids are all signed integers except in the
libwacom_match code where they are uint32_t. We can't change that
without API changes (not ABI though), so let's just document it and move on.


skomra:
Reviewed-by: Aaron Armstrong Skomra <aaron.skomra@wacom.com>

PR #110, opened 2019-06-25
Add support for stylus groups

### LINKED ISSUE #99: Group pens into specific sets
See #57 for the need/reasoning.

Basic idea: instead of each tablet listing every possible serial number, group those into various sets, e.g. "Pro Pen" or "AES Pen". Then add those sets to the tablets.

Advantages:
- easier to identify if we forgot to add a specific pen to a tablet
- easier to update the serials when they change (e.g. some tablets support multiple generations of the same pen)

The first big TODO item here is to identify the current serial numbers and identify which ones could be grouped by sets.

whot:
Fixes #99 

cc @skomra


whot:
fwiw, I'm just guessing at stylus group names here. If you have better knowledge, code-names, or otherwise group names that are better than the one I picked, I'm all ears :)

This is internal API, so we can change this in the future too.

whot:
Ok, all but the `0xfffff?` styli are in a group now. Not 100% whether the groups are correct or whether you want to split them up in some other way, but I'd appreciate a check.

diff from `list-compatible-styli` before and after shows identical data

skomra:
I can't vouch for the historical accuracy of the group names, but they help make it easy to form a mental model of what's going on with the different groups. 

This is just a nit that doesn't need to be fixed, but just fyi some Styli lines end with a semicolon and some don't. Example:

Styli=@intuos3-pucks;@intuos3 #5 instances
Styli=@intuos3-pucks;@intuos3; # intuos3-12x12.tablet:40:

I think previously (almost) all lines ended with a semicolon.

Reviewed-by: Aaron Armstrong Skomra <aaron.skomra@wacom.com>

whot:
Fixed and appended the semicolons now, thanks for the review!

PR #113, opened 2019-06-27
data: move the git and svg checks to a separate scripts

whot:
Easier to test, comment, write, and review when it's in a separate file.

The automake integration is a bit odd. We can't pass arguments to scripts but
the only sane way for that script is to take the top_srcdir as argument
(otherwise it'd have to be a configure.in script). We can't do that, but we
can set it in the environment for the tests so that's the next best thing.

This also drops the custom lookups for git, grep and sed. Just make sure it's
in your PATH. Where git ls-files fails, the test is skipped which should make
anyone curious. This also covers the case where git is not available, so
that's not a hard failure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ stamp-h1
 *.la
 *.swp
 *.trs
+*.log
 .deps
 .libs
 test-driver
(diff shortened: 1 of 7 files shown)


whot:
sorry about the churn, test case failure is fixed now by switching to bash as the script interpreter (as opposed to `sh`)

PR #120, opened 2019-07-02
Hide the accidentally exposed API

### LINKED ISSUE #114: Drop the accidentally exposed API
API that was exported in the past because it was scooped up by the symbol regex (removed in e4ef4168aeb23b6e375999a820c1dc453d138920) includes:
```
libwacom_error_set()
libwacom_stylus_destroy()
libwacom_update_match()
libwacom_match_new()
libwacom_match_destroy()
```

Now that we have proper symbol handling (43981e9d70a80d42b873f46c32bfe66f7a413619) we should be able to drop the above from the API but still retain ABI-compatibility with older versions.

I don't expect anyone to ever have used those, so while it may be an API break, it's not really one given that all callers will continue to compile fine as-is.

whot:
We have a few functions that ended up in the ABI despite being intended as internal-only:
```
libwacom_error_set()
libwacom_stylus_destroy()
libwacom_update_match()
libwacom_match_new()
libwacom_match_destroy()
```
They simply got scooped up by the `libwacom_*` regex (removed in 43981e9d70a80d42b873f46c32bfe66f7a413619).

Of these functions, only `error_set` and `match_new` are safe to call. `update_match` messes with the database, the two `destroy` functions would cause double-frees if called directly. None of these have every been in a public header file, so to invoke any of them you'd have to either a) `dlsym` or b) declare it locally. Neither of which makes sense given what the functions do. It's safe to say that there are 0 legitimate users out there.

Nonetheless, I have an interest in not breaking the ABI :) at least initially anyway.

This patchset
- replaces most of the functions with internal-only ones for internal use
- makes those functions noops
- binds them to the current symbol version only through aliases, thus clearing up the namespace
- restores *some* of the function names for internal-only use (simply because those names are the most sensible ones)

For anyone *not* using these functions (i.e. 100% of our users), nothing changes whatsoever.

For anyone who *does* use any of these functions (and really shouldn't):
- a binary compiled against libwacom 0.33 and earlier will continue to work
- recompiling against this version now will result in an unresolved symbol for any of those functions above

This is a lot of effort for what we should really just fix with a soname bump and complete removal of those functions. But that *requires* a rebuild of *all* users. So until the soname bump happens, this is an ABI-compatible solution for, say, enterprise distributions that have an interest in not breaking the ABI :)

Fixes #114


whot:
Test script is [here](https://gist.github.com/whot/41413074983b25a2e14f8b9310ca572e), just make sure the source directories are on the right shas (`libwacom-0.33` and `master`).

And to compare the binary output for the `.so` before and after:
```
$ diff -u \
    <(nm libwacom-inst-0.33/lib/libwacom.so.2.6.1| grep "T" | cut -d' ' -f3) \
    <(nm libwacom-inst-0.34/lib/libwacom.so.2.6.1 |grep "T" | cut -d' ' -f3)
```

whot:
oh, sorry, that was just a mistake. I did intend to just restore it as `libwacom_error_set()` as it was before, but this time internal-only. I've replaced this now. thanks for the review

whot:
doh. I forgot to push before hitting the merge button. This is pushed as follow-up now, see 1da3dba8f1959976a74c425292a5f1fa01c9d2e9

Issue #103, opened 2019-06-24
show-svg-image is broken


whot:
Fedora 30, this seems to only show a black screen. @jigpu, is this still working for you? Last comit was from you ~3y ago :)

PR #121, opened 2019-07-02
Prevent librsvg from bailing at the sight of uninitialized data

jigpu:
Leaving 'error' uninitalized causes the librsvg to believe that Something
Bad (TM) has already happened and prevents it from working.

(lt-show-svg-image:18442): librsvg-CRITICAL **: 14:34:08.414: rsvg_handle_write: assertion 'error == NULL || *error == NULL' failed

Fixes: https://github.com/linuxwacom/libwacom/issues/103

### DIFF
--- a/tools/show-svg-image.c
+++ b/tools/show-svg-image.c
@@ -209,6 +209,8 @@ update_tablet (Tablet *tablet)
 	gchar       *file_data, *escaped_file_data, *data;
 	gsize        file_len;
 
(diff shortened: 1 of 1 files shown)


jigpu:
I'm out of the office for the next week so it'll be a bit before I can make changes. FWIW I was following the style of `main()`

whot:
eh, good enough, not worth it to wait then :)

PR #122, opened 2019-07-03
test: switch to using glib test suites

whot:
The current tests just use `assert()`. Which is only fine for local debugging where one executes the test directly. Almost all the time I need to run gdb to figure out what's going on and which device is actually broken.

Fix this by using the glib test suite instead. This makes each test a lot more self-contained and quicker to understand than the previous mess. And the output is a lot more useful now - example from `test-load` with a forced failure:
```
/load/0000:0000: OK
/load/056a:00bc: OK
/load/056a:00b8: OK
/load/056a:00f4: **
ERROR:test-load.c:160:test_cintiq24hd: assertion failed (libwacom_get_ring_num_modes(device) == 4): (3 == 4)
FAIL
/load/056a:00cc: OK
/load/056a:00f8: OK
/load/056a:0304: OK
/load/056a:0065: OK
/load/056a:WACf004: OK
```

Because we already use glib, the test suite doesn't introduce a new dependency.


test/test-stylus-validity.c
```diff
@@ -0,0 +1,300 @@
...
+
+	if (libwacom_stylus_is_eraser(stylus))
+		add_test(stylus, test_buttons);
+
+	if (libwacom_stylus_get_type(stylus) != WSTYLUS_PUCK) {
+		switch (libwacom_stylus_get_id(stylus)) {
+			case 0xffffd:
+			case 0x8e2:
+			case 0x862:
+				break;
```
jigpu: It should be more straightforward to check 0x8e2 and 0x862 for pressure/distance now that this is a switch statement. Don't have to if you don't want though.

whot: Going to leave that as follow-up, mostly because... huh? Sorry, didn't quite parse what you meant :)

whot: Included in #132

test/test-stylus-validity.c
```diff
@@ -0,0 +1,300 @@
...
+	add_test(stylus, test_type);
+
+	if (libwacom_stylus_has_eraser(stylus))
+		add_test(stylus, test_eraser);
+
+	if (libwacom_stylus_get_type(stylus) == WSTYLUS_PUCK)
+		add_test(stylus, test_puck);
+
+	if (libwacom_stylus_is_eraser(stylus))
+		add_test(stylus, test_buttons);
```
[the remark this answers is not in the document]
whot: Included in #132

PR #124, opened 2019-07-04
Add distcheck and builddirs to travis

whot:
Run `make distcheck` in travis to pick up on those breakages.

And do the whole thing in a custom builddir so `srcdir` != `builddir` because that's a common source of issues.


whot:
@jigpu how do I check whether the coverity script still works with this?

whot:
This conflicts with #125 so if you're happy with that one, we can just drop this one. Sorry about the mess.

PR #125, opened 2019-07-04
travis CI: simplify the travis script a bit

whot:
This started out as something more complicated but for now just this set will do and already helps a bit.

Conflicts with #124 but if we're happy with this we can just close the other one - it includes the changes here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: required
+dist: xenial
 language: c
 
 compiler:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,24 +12,42 @@ addons:
       - libgudev-1.0-dev
       - libxml++2.6-dev
 
-before_install:
-  - echo -n | openssl s_client -connect scan.coverity.com:443 | sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p' | sudo tee -a /etc/ssl/certs/ca-
+stages:
```
jigpu: Travis apparently tries to run the Coverity stage prior to the default "test" stage unless you explicitly list `  - test` at the beginning of the stages list. No need to bother Coverity if the compile or unit tests fail.

whot:
Merging this based on the one nitpick being fixed. If there's more we can fix it up as follow-up. Thanks

PR #134, opened 2019-07-19
travis CI: remove a now-obsolete condition

whot:
This line is only executed in the coverity bits, so that string is always
nonzero and we don't need to test for it


whot:
@jigpu I'll need you to test this since it's coverity. Thanks.

PR #141, opened 2019-08-05
Refcount cleanups

whot:


libwacom/libwacom-database.c
```diff
@@ -736,13 +738,17 @@ libwacom_parse_tablet_keyfile(WacomDeviceDatabase *db,
...
+
 out:
 	if (path)
 		g_free(path);
 	if (keyfile)
 		g_key_file_free(keyfile);
 	if (error)
 		g_error_free(error);
+	if (!success)
+		device = libwacom_unref(device);
```
jigpu: Might also remove the `g_free (device); device = NULL` statements up on line 602 (the "failed to match" error) since this will take care of it too?

whot: fixed, thanks. together with an indentation fix in that same spot.

PR #145, opened 2019-08-07
Add meson_options.txt to EXTRA_DIST

whot:
And add the required bits to the test systems so this can't happen again.


whot:
new record for typos in one commit...

PR #147, opened 2019-08-08
travis ci: build with autotools from a tarball too

whot:
Picks up missing EXTRA_DIST entries and similar.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ jobs:
         - pushd builddir > /dev/null
         - meson test --setup=valgrind --suite=valgrind || (cat meson-logs/testlog-valgrind.txt && false)
         - popd > /dev/null
-    #  build from a tarball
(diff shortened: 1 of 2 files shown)


whot:
> The tarball should include the configure script, shouldn't it?

That is true only as long as we build it with autotools, a meason dist wouldn't have it. (yeah, I know, distant future, but nonetheless)

That aside, at least in fedora we're building the package with `autoreconf` before running `configure` and I think other distros do the same. For historical reasons that ended up being more reliable than just running the configure script, at least for distribution building.

It's "free" to add another stage, so we can just do that I guess.

whot:
Updated to include a pure `configure` stage from the tarball.

And another commit on top to `make install` from the tarball and run `tree` on it. This won't pick up any errors unless `make install` itself fails, but it'll give us the log of what files get installed which may come in handy one day.

PR #148, opened 2019-08-08
Install the udev rules

### LINKED ISSUE #144: Install the generated udev rules as part of the install process
Pretty sure anyone who needs to install a new libwacom also needs to update the udev rules to ensure the correct `ID_INPUT_TABLET`/`ID_INPUT_TABLET_PAD` tagging. Without that we may have issues where users manually install but then the pad isn't tagged correctly (this would affect libinput, not the xf86-input-wacom driver).

So let's fix this by just installing the udev rules, they're there for a reason after all.

whot:
Fixes #144

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ jobs:
         - mkdir _build
         - pushd _build > /dev/null
         - ../autogen.sh --disable-silent-rules
-        - make dist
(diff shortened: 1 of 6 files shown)


whot:
Fixed the commit message to refer to the udev base directory instead (with examples), thanks

PR #158, opened 2019-09-18
Add support for ELAN 2628

MusiKid:
Hi!
This tablet is also found on some HP Envy x360 ag0xxxx (including mine). This commit is basically just a copy of  #95 (thanks @daenney !) modified to match this device.


MusiKid:
Oh sorry I didn't know about this! 
I will add it ASAP.

MusiKid:
The test is passed now, so I think the only problem was that I didn't add the definition to the `meson.build`.

whot:
yep, thanks!

Issue #162, opened 2019-09-21
pkg-config file is inconsistent between autotools and meson.


jchv:
As @worldofpeace has noted in https://github.com/NixOS/nixpkgs/pull/69182#issuecomment-533764724, the pkg-config file generated by Meson differs from the pkg-config file generated by autotools. Here is a manually normalized diff showing the actual differences:

```diff
 prefix=/nix/store/xxx-libwacom-1.1
-exec_prefix=${prefix}
-libdir=${exec_prefix}/lib
+libdir=${prefix}/lib
 includedir=${prefix}/include
 
-Name: libwacom
+Name: Libwacom
 Description: Wacom model feature query library
 Version: 1.1
-Requires.private: glib-2.0
 Cflags: -I${includedir}/libwacom-1.0
 Libs: -L${libdir} -lwacom
```

I think the `exec_prefix` bits are benign, but it looks like the `Name` and `Requires.private` changes were not intended.

I believe the only realistic side-effect here is that behavior when attempting to statically link may change.

whot:
thanks! I've pushed the fix for the `Name` bit now too. Out of interest: does any machine care about the `Name`? humans won't care about capitalization I suspect.

jchv:
I don't think that any problems were caused by the name change; I didn't notice any failures caused by that, at least.

worldofpeace:
I think it's good to keep things consistent. I do believe that `Name` is for humans though.

PR #163, opened 2019-09-21
meson: add private requires to .pc

worldofpeace:
Fixes https://github.com/linuxwacom/libwacom/issues/162


worldofpeace:
This will probably add `libgudev` as well.

PR #171, opened 2019-10-27
meson: make docs configurable

okias:
Nice to have for distributions, which needs to have predicable build.

Signed-off-by: David Heidelberg <david@ixit.cz>


meson_options.txt
```diff
@@ -1,3 +1,7 @@
+option('doc',
```
whot: let's make this `documentation`, immediately more obvious and most users will only ever have to type this once anyway.

PR #174, opened 2019-11-04
meson.build: disable the deprecated symbols test with LTO

### LINKED ISSUE #152: 1.0: libwacom is not LTO ready
When compiled with LTO optimisation linkiong of the test suite fails:

<pre>[tkloczko@barrel test]$ make
/bin/sh ../libtool  --tag=CC   --mode=link gcc  -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -flto -Os -no-install -Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -flto -fuse-linker-plugin -o test-deprecated test-deprecated.o ../libwacom/libwacom.la  -ldl 
libtool: link: gcc -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -flto -Os -Wl,-z -Wl,relro -Wl,--as-needed -Wl,-z -Wl,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -flto -fuse-linker-plugin -o test-deprecated test-deprecated.o  ../libwacom/.libs/libwacom.so -lgudev-1.0 -lgobject-2.0 -lglib-2.0 -ldl -Wl,-rpath -Wl,/home/tkloczko/rpmbuild/BUILD/libwacom-0.99.901/libwacom/.libs
/usr/bin/ld: /tmp/test-deprecated.thrYTc.ltrans0.ltrans.o: in function `main&apos;:
/home/tkloczko/rpmbuild/BUILD/libwacom-0.99.901/test/test-deprecated.c:77: undefined reference to `libwacom_match_destroy@LIBWACOM_0.33&apos;
/usr/bin/ld: /home/tkloczko/rpmbuild/BUILD/libwacom-0.99.901/test/test-deprecated.c:78: undefined reference to `libwacom_match_new@LIBWACOM_0.33&apos;
/usr/bin/ld: /home/tkloczko/rpmbuild/BUILD/libwacom-0.99.901/test/test-deprecated.c:79: undefined reference to `libwacom_error_set@LIBWACOM_0.33&apos;
/usr/bin/ld: /home/tkloczko/rpmbuild/BUILD/libwacom-0.99.901/test/test-deprecated.c:80: undefined reference to `libwacom_stylus_destroy@LIBWACOM_0.33&apos;
/usr/bin/ld: /home/tkloczko/rpmbuild/BUILD/libwacom-0.99.901/test/test-deprecated.c:81: undefined reference to `libwacom_update_match@LIBWACOM_0.33&apos;
collect2: error: ld returned 1 exit status
make: *** [Makefile:664: test-deprecated] Error 1
make: Target &apos;all&apos; not remade because of errors.
</pre>

whot:
If I understand this correctly:
- our deprecated symbols are remapped into a specific library version
- nothing in the test calls those symbols, so they're optimized away (they end
  up as local symbols)
- our test uses asm to link to those functions but by then they're already
  gone

Rather than fixing this let's just disable the test, it doesn't provide us
with anything useful in the LTO condition anyway.

Fixes #152

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -543,13 +543,18 @@ if get_option('tests')
 		message('valgrind not found, disabling valgrind test suite')
 	endif
 
-	test_deprecated = executable('test-deprecated',
-				     'test/test-deprecated.c',
-				     dependencies: [dep_libwacom, dep_dl],
-				     include_directories: [includes_src],
-				     c_args: tests_cflags,
-				     install: false)
-	test('test-deprecated', test_deprecated, suite: ['all'])
+	# because of the tricks we use in calling the function and exposing
+	# it to begin with, LTO gets confused and this fails to link.
+	# Let's just disable it here.
+	if not get_option('b_lto')
+		test_deprecated = executable('test-deprecated',
+					     'test/test-deprecated.c',
+					     dependencies: [dep_libwacom, dep_dl],
+					     include_directories: [includes_src],
+					     c_args: tests_cflags,
+					     install: false)
+		test('test-deprecated', test_deprecated, suite: ['all'])
+	endif
 endif
 
 # This is a non-optional test


kloczek:
This solution is kind of wrong. Why? Because assumes that someone will be using LTO by specify in meson options -D b_lto=enable.

In reality it is possible to use LTO with meson completely transparently by passing all necessary options in AR, NM, RANLIB, CFLAGS and LDFLAGS environment variables.

Other thing is that implementing LTO in this PR way assumes that someone will be using gcc. Build framework should not be assuming use any any type of the compiler. All what is necessary to do here is possible to pass over above env variables.

IMO it should be only meson option to disable those tests if there is no other way to fix that issue.

whot:
it's always possible to pass/modify/hack the build process with various scripts and environment variables. That's a moving target we cannot maintain though. Plus, if you control the environment to that level in your build process, you can also patch the code to change what's being built.

We already have an option to disable tests (`-Dtests=false`), adding another one just for one specific test to accommodate for a fairly specialised build method is not a good idea.

kloczek:
Correct me if I'm wrong AFAIK -Dtests=false disables all tests or only those affected by LTO.

Still I'm not sure what is exact cause that that test usnit is failing with LTO. gcc/ld bug? Because if it is like this dicussing that issue here is a bit pointless.

whot:
> Correct me if I'm wrong AFAIK -Dtests=false disables all tests or only those affected by LTO.

Correct, it disables everything.

> Still I'm not sure what is exact cause that that test usnit is failing with LTO

Have a look at the [source of the test](https://github.com/linuxwacom/libwacom/blob/master/test/test-deprecated.c). The test:
- `dlopen`'s `libwacom.so`
- tries to `dlsym` the deprecated functions which *should* fail
- then it calls the functions directly because we've got the `asm` statements for the right library version.

So it's a hacky test but it works. Whatever LTO does exactly, it seems to ignore the `asm .symver` statements, making those calls fail with linker errors. I don't know enough though to go further here though.

PR #178, opened 2019-11-05
Add AES pen IDs

jigpu:
Adds AES pen IDs to the stylus database as well as some additional infrastructure to make them a bit easier to work with. Integrated #177

### DIFF
--- a/data/libwacom.stylus
+++ b/data/libwacom.stylus
@@ -20,6 +20,204 @@ Buttons=2
 Axes=Pressure;
 Type=General
 
+[0x1]
+# Lenovo ; VID_NONE     | 0x0000 | BAT_SWAP
+Name=AES Pen
+Group=isdv4-aes
+Buttons=1
+IsEraser=true
+Axes=Pressure
+Type=Mobile
+
+[0x11]
+# Lenovo ; VID_NONE     | 0x0000 | BAT_CHRG
+Name=AES Pen
+Group=isdv4-aes
+Buttons=1
+IsEraser=true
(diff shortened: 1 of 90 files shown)


data/libwacom.stylus
```diff
@@ -20,6 +20,204 @@ Buttons=2
...
+IsEraser=true
+Axes=Pressure
+Type=Mobile
+
+[0x11]
+# Lenovo ; VID_NONE     | 0x0000 | BAT_CHRG
+Name=AES Pen
+Group=isdv4-aes
+Buttons=1
+IsEraser=true
```
whot: I don't think `IsEraser` is correct here because that is based on the assumption that the pen is **always** an eraser (like the eraser tip on the other pens). We need some other toggle here for the eraser button.

jigpu: Agreed. It seems like the code implies that `IsEraser` is _always_ an eraser, and `HasEraser` is _never_ an eraser. I basically flipped a coin on which to use and figured a proper solution would be worked out in #179.

Looking at how GNOME uses it, `HasEraser` may be the slightly better choice...

whot: I'd go with either waiting for #179 or leaving it as false for both right now. I'd lean towards the former, less work overall I reckon.

binli:
> Hi @binli -- this commit is part of an open pull request which has not (yet) been merged. Watch #178 for updates. It and the eraser button patches should be merged soon.
Jigpu,
 Thanks for the information. :)

PR #179, opened 2019-11-05
WIP: Modifications to libwacom eraser handling

jigpu:
I'm not totally happy with how erasers are expressed for AES pens, so I've been toying around with some ideas... Not ready for prime-time, but maybe ready for light comments :)

On top of #177 and #178

### DIFF
--- a/data/libwacom.stylus
+++ b/data/libwacom.stylus
@@ -1,13 +1,14 @@
 # Some generic fallback styli
 [0xfffff]
 Name=General Pen
+PairedStylusIds=0xffffe;
 Buttons=2
-HasEraser=true
 Axes=Tilt;Pressure;Distance;
 Type=General
 
 [0xffffe]
 Name=General Pen Eraser
+PairedStylusIds=0xfffff;
 EraserType=Invert
 Buttons=2
 Axes=Tilt;Pressure;Distance;
@@ -15,7 +16,6 @@ Type=General
 
 [0xffffd]
 Name=General Pen with no Eraser
-HasEraser=false
 Buttons=2
 Axes=Pressure;
 Type=General
@@ -224,7 +224,6 @@ Type=Mobile
 Name=Inking Pen
 Group=intuos
 Buttons=0
-HasEraser=false
 Axes=Tilt;Pressure;Distance;
 Type=Inking
 
@@ -233,7 +232,6 @@ Type=Inking
 Name=Inking Pen
 Group=intuos3
 Buttons=0
-HasEraser=false
 Axes=Tilt;Pressure;Distance;
 Type=Inking
 
@@ -241,7 +239,6 @@ Type=Inking
 # Intuos4, 5 and Cintiq 21UX2, 22HD, 24HD
 Name=Inking Pen
 Buttons=0
-HasEraser=false
 Axes=Tilt;Pressure;Distance;
 Type=Inking
 
@@ -250,7 +247,6 @@ Type=Inking
 Name=Inking Pen
 Group=intuos5
 Buttons=0
-HasEraser=false
 Axes=Tilt;Pressure;Distance;
 Type=Inking
 
@@ -259,7 +255,7 @@ Type=Inking
 # Intuos and Intuos2
 Name=Classic Pen
 Group=intuos
-HasEraser=true
+PairedStylusIds=0x82a;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=Classic
@@ -268,7 +264,6 @@ Type=Classic
 # Intuos P/PT 2
 Name=Intuos Pen
 Group=intuospt
-HasEraser=false
 Buttons=2
 Axes=Pressure;Distance;
 Type=General
@@ -277,7 +272,6 @@ Type=General
 # Intuos P/PT 2
 Name=Intuos Pen
 Group=intuospt3
-HasEraser=false
 Buttons=2
 Axes=Pressure;Distance;
 Type=General
@@ -286,13 +280,13 @@ Type=General
 # [0x842]
 # Intuos2
 # Name=Designer Pen
-# HasEraser=false
+# PairedStylusIds=0x84a;
 
 [0x842]
 # MobileStudio Pro
 Name=Pro Pen
 Group=mobilestudio
-HasEraser=true
+PairedStylusIds=0x84a;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=General
@@ -301,7 +295,6 @@ Type=General
 # MobileStudio Pro
 Name=Pro Pen 3D
 Group=mobilestudio
-HasEraser=false
 Buttons=3
 Axes=Tilt;Pressure;Distance;
 Type=3D
@@ -310,7 +303,7 @@ Type=3D
 # Intuos2
 Name=Grip Pen
 Group=intuos2
-HasEraser=true
+PairedStylusIds=0x85a;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=General
@@ -319,7 +312,7 @@ Type=General
 # Intuos3 and Cintiq 21UX
 Name=Grip Pen
 Group=intuos3
-HasEraser=true
+PairedStylusIds=0x82b;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=General
@@ -328,7 +321,7 @@ Type=General
 # Intuos3 and Cintiq 21UX
 Name=Classic Pen
 Group=intuos3
-HasEraser=true
+PairedStylusIds=0x81b;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=Classic
@@ -337,7 +330,6 @@ Type=Classic
 # Intuos3 and Cintiq 21UX
 Name=6D Art Pen
 Group=intuos3
-HasEraser=false
 Buttons=0
 Axes=Tilt;Pressure;Distance;RotationZ;
 Type=Marker
@@ -346,7 +338,7 @@ Type=Marker
 # Intuos4, 5 and Cintiq 21UX2, 24HD
 Name=Grip Pen
 Group=intuos4
-HasEraser=true
+PairedStylusIds=0x80a;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=General
@@ -355,7 +347,7 @@ Type=General
 # Intuos4, 5 and Cintiq 21UX2, 24HD
 Name=Art Pen
 Group=intuos4
-HasEraser=true
+PairedStylusIds=0x80c;
 Buttons=2
 Axes=Tilt;Pressure;Distance;RotationZ;
 Type=Marker
@@ -364,7 +356,7 @@ Type=Marker
 # Intuos4, 5 and Cintiq 21UX2, 22HD, 24HD
 Name=Art Pen
 Group=intuos5
-HasEraser=true
+PairedStylusIds=0x10080c;
 Buttons=2
 Axes=Tilt;Pressure;Distance;RotationZ;
 Type=Marker
@@ -373,7 +365,7 @@ Type=Marker
 # Intuos4, 5 and Cintiq 21UX2, 24HD
 Name=Grip Pen
 Group=intuos5
-HasEraser=true
+PairedStylusIds=0x10080a;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=General
@@ -382,7 +374,7 @@ Type=General
 # MobileStudio Pro, Cintiq Pro, Intuos Pro
 Name=Pro Pen Slim
 Group=mobilestudio
-HasEraser=true
+PairedStylusIds=0x10084a;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=General
@@ -390,7 +382,7 @@ Type=General
 [0x40802]
 # Intuos4, 5 and Cintiq 21UX2, 24HD
 Name=Classic Pen
-HasEraser=true
+PairedStylusIds=0x4080a;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=Classic
@@ -399,7 +391,7 @@ Type=Classic
 # Intuos4, 5 and Cintiq 21UX2, 24HD
 Name=Classic Pen
 Group=intuos5
-HasEraser=true
+PairedStylusIds=0x14080a;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=Classic
@@ -408,7 +400,7 @@ Type=Classic
 # Cintiq 13HD Pro Pen
 Name=Pro Pen
 Group=intuos5
-HasEraser=true
+PairedStylusIds=0x16080a;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=Classic
@@ -417,7 +409,7 @@ Type=Classic
 # DTH2242 Pen
 Name=Pen
 Group=dth2242
-HasEraser=true
+PairedStylusIds=0x18080a;
 Buttons=2
 Axes=Tilt;Pressure;Distance;
 Type=General
@@ -427,7 +419,6 @@ Type=General
 #Intuos and Intuos2
 Name=Stroke Pen
 Group=intuos
-HasEraser=false
 Buttons=0
 Axes=Tilt;Pressure;Distance;
 Type=Stroke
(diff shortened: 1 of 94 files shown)


libwacom/libwacom.c
```diff
@@ -1225,7 +1225,7 @@ libwacom_stylus_has_eraser (const WacomStylus *stylus)
 LIBWACOM_EXPORT int
 libwacom_stylus_is_eraser (const WacomStylus *stylus)
 {
-	return stylus->is_eraser;
+	return libwacom_stylus_get_eraser_type(stylus) != WACOM_ERASER_NONE;
```
whot: I think this should only return true where the tool itself is only ever an eraser, i.e. for `WACOM_ERASER_INVERT`. That seems to be the most reliable backwards-compatibility map.

libwacom/libwacom.h
```diff
@@ -638,6 +648,12 @@ WacomAxisTypeFlags libwacom_stylus_get_axes (const WacomStylus *stylus);
  */
 WacomStylusType libwacom_stylus_get_type (const WacomStylus *stylus);
 
+/**
+ * @param stylus The stylus to query
+ * @return The type of eraser on the stylus
+ */
+WacomEraserType libwacom_stylus_get_eraser_type (const WacomStylus *stylus);
```
whot: I'm surprised this builds. You need to add this to the `libwacom.sym` files like this:
```
LIBWACOM_1.2 {
   libwacom_stylus_get_eraser_type;
} LIBWACOM_0.33;
```

data/libwacom.stylus
```diff
@@ -1,21 +1,21 @@
 # Some generic fallback styli
 [0xfffff]
 Name=General Pen
+PairedIds=0xffffe;
```
whot: my bikeshed here is that we should use a different name. `PairedId` is already used for tablet files and I'd rather not have any accidental overlap here. 

"well, what's a better name then", you ask? true, you got me there... `PairedStylusId` at least to have the clear differentiation.

libwacom/libwacom.c
```diff
@@ -1206,6 +1206,18 @@ libwacom_stylus_get_name (const WacomStylus *stylus)
...
+{
+	return stylus->num_ids;
+}
+
+LIBWACOM_EXPORT const int *
+libwacom_stylus_get_paired_ids(const WacomStylus *stylus)
+{
+	return stylus->paired_ids;
+}
+
```
whot: IMO better:
```c
const int *libwacom_stylus_get_paired_ids(const WacomDevice *device, int *num_paired_ids);
```
That's the prototype we use for `libwacom_get_supported_styli` already. And I can't think of a use-case where we only need the num_paired anyway, so having just one function here is better.

libwacom/libwacom.sym
```diff
@@ -61,6 +61,8 @@ global:
     libwacom_stylus_get_id;
     libwacom_stylus_get_name;
     libwacom_stylus_get_num_buttons;
+    libwacom_stylus_get_num_paired_ids;
+    libwacom_stylus_get_paired_ids;
```
whot: haha, found the right emoji: :rotating_light: :rotating_light: :rotating_light: 

this must go into a new section, otherwise you're changing the ABI. Basically: once a release is out, the existing sections in `libwacom.sym` must never be touched and any new additions go into the new section. See my other comment for how the section would look like, you can add those two there.

libwacom/libwacom-database.c
```diff
@@ -266,25 +266,38 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
+		if (string_list) {
+			GArray *array;
+			guint i;
+
+			array = g_array_new (FALSE, FALSE, sizeof(int));
+			for (i = 0; string_list[i]; i++) {
+				char *endptr;
+				long val;
+
+				val = strtol (string_list[i], &endptr, 0);
```
whot: probably best to copy `safe_atoi` from libinput for this? There's also `g_ascii_string_to_unsigned()` but it doesn't handle hex properly.

libwacom/libwacom-database.c
```diff
@@ -336,6 +349,34 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
+		const int* ids = libwacom_stylus_get_paired_ids(stylus);
+		int count = libwacom_stylus_get_num_paired_ids(stylus);
+		int i;
+
+		for (i = 0; i < count; i++) {
+			const WacomStylus *pair;
+
+			pair = libwacom_stylus_get_for_id(db, ids[i]);
+			if (pair == NULL) {
+				g_warning("Paired stylus '0x%x' not found, ignoring.", ids[i]);
```
whot: fwiw, printf directive `%#x` prints a hex number prefixed with `0x`

libwacom/libwacom.h
jigpu: Looks like my update to that file didn't get added into git :man_facepalming:

libwacom/libwacom.sym
jigpu: Thanks for the tip! Wasn't quite sure how to deal with the ABI change in this new symbol-versioned reality :+1:

data/libwacom.stylus
jigpu: ACK

libwacom/libwacom.c
jigpu: ACK

libwacom/libwacom-database.c
jigpu: :+1:

libwacom/libwacom-database.c
jigpu: Looks like `safe_atoi_base` could handle the hex IDs, though with the minor caveat at the IDs would then have to always be specified in hex. I suppose we could also add "0" to the list of accepted bases, but I don't think we get much from allowing both decimal and hex.

libwacom/libwacom.c
jigpu: That sounds fine to me. How do we want `libwacom_stylus_has_eraser` to behave? Right now it only returns true if a stylus is paired with an eraser, but is not itself one. Do we want it to also return true if the stylus may _become_ an eraser (i.e., is an AES pen)? Alternatively, do we want to introduce a new function (e.g. `libwacom_stylus_can_erase`) that would be true for an AES pen?

Whatever we do, those functions could use some documentation that clarifies the difference between "is" and "has" (and possibly "can") :wink:

whot: I'd say exact same behaviour as now (i.e. return `false`) and deprecate it. A quick check at the gnome-control-center shows that the `has-eraser` flag is used in combination with the button count to determine the "layout" of the panel. from libwacom's POV that is outside our control so the safest thing is to not break anything.

gnome is fast-moving enough that this won't be an issue for long anyway.


> do we want to introduce a new function (e.g. libwacom_stylus_can_erase)

yes, but why not go for the same approach with `get_eraser_type()` and return an enum value? More future-proof IMO.

libwacom/libwacom-database.c
whot: I would cheer on any CI patches that check for the pen IDs to always be in `0xabcd` format. No point having this part-decimal, part-hex anyway.

libwacom/libwacom-database.c
```diff
@@ -246,29 +262,41 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
+			for (i = 0; string_list[i]; i++) {
+				char *endptr;
+				long val;
+
+				val = strtol (string_list[i], &endptr, 0);
+				if (*endptr == '\0') {
+					g_array_append_val (array, val);
+					stylus->num_ids++;
+				} else {
+					g_warning ("Stylus %s (%s) %s\n", stylus->name, groups[i], "Ignoring invalid PairedId value");
```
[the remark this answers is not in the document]
jigpu: ACK :)

libwacom/libwacom-database.c
jigpu: New commit added for this.

libwacom/libwacom.c
jigpu: The more I look at this, the more I think compatibility really isn't an issue and the more I prefer the implementation under review. As far as I can tell, the behavior _is_ backwards compatible:

* Callers expecting the current behavior will continue to function correctly with the various invert-style erasers that we may introduce in the future.
* Callers expecting future behavior will work correctly with current and past versions of libwacom for the styli that are in their databases (which are all invert-style).

The only compatibility argument that I can see is that callers which expect current behavior may mistreat the new button-style erasers as invert-style. I'm not concerned about this, however, since those callers would need other updates (even _if_ we changed the API) to properly support button-style erasers.

jigpu:
Most issues should be resolved, so I'm requesting a second round of reviews

libwacom/libwacom.c
whot: unless I'm misunderstanding something, my worry is about *current callers* with *updated libwacom*. Right now, we have pens that **have** an eraser and pens that **are** an eraser (but those have a different ID). In the case of AES pens, I don't think they change the ID for the eraser, so with the check for `!= NONE` we now basically say that every AES pen is an eraser pen and there are no stylus pens anymore.

I still think it'd be better to fail the other way around, so older applications will think every AES pen is a stylus without an eraser.

libwacom/libwacom-database.c
```diff
@@ -28,7 +28,9 @@
 #include "config.h"
 #endif
 
+#define _GNU_SOURCE 1
```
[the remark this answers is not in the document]
jigpu: Fixed

libwacom/libwacom.c
jigpu: It seems just as wrong for an application to treat an AES pen as a having no eraser as it is to treat the pen as *being* the eraser. Your AES pen goes from "doesn't work at all" with the current libwacom release to "works incorrectly" either way. The only way for an application to get to "works correctly" is to start using `libwacom_stylus_get_eraser_type` and write new code to handle button-style erasers.

The reason I prefer `!= NONE` comes down to trying to find an eraser given an arbitrary tool ID. With this definition I can simply call `is_eraser` on each tool and know when I've found an ID that can erase. Using `== WACOM_ERASER_INVERT` instead makes `is_eraser` useless for me, and so I end up write my own `!= NONE` check anyway.

jigpu:
The commit list shown here seems to be sorted by date rather than the actual topological order... I wonder if it was intended to make it easier to see which patches have changed since a given date? The topological order does have the "Mobile" commit before the commit to add the AES IDs.

libwacom/libwacom.c
whot: both are wrong, but I'd say one is less so than the other. not having a working eraser button should be fine for a large fraction of use-cases [1] esp given that there are also calls for mapping the eraser button to just be a stylus button. So right now, having the eraser button not work doesn't bother me as much as having the stylus potentially not working.

> Using `== WACOM_ERASER_INVERT` instead makes is_eraser useless for me, and so Iend up write my own != NONE check anyway.

that's true, but only for the simplest client. If we assume that a client is interested in whether an eraser is a button vs pen, then that doesn't apply anymore because you need a custom check anyway. And you'd be using  `libwacom_stylus_get_eraser_type` instead.
Note that libwacom isn't a framework, we have what - 2 or 3 actual users with little to no churn in the code that calls our API. So even if it is a few more lines for those they get written once and then never touched again.


[1] and we're just talking about backwards-compat anyway.

jigpu: Sorry for the slow rate of replies. Been bouncing between a bunch of different tasks for the past several weeks :(

I'm still not really buying this argument. You might not be able to configure your AES pen because the tool thinks its an eraser, but its still going to work. The worst case scenario I can think of is a caller monitoring the event stream and using the result of `is_eraser` (rather than the event's own stylus/eraser flag) to tell applications "you should erase with this tool". None of our "2 or 3 actual users" have ever done done this, and I doubt there are any others out there that do. Libwacom is primarily used for configuration after all. In reality you just won't be able to properly configure your AES pen, which isn't any different from today.

> that's true, but only for the simplest client. If we assume that a client is interested in whether an eraser is a button vs pen, then that doesn't apply anymore because you need a custom check anyway. And you'd be using libwacom_stylus_get_eraser_type instead.

I for one found it handy to have them around as utility functions within libwacom itself. I imagine you wouldn't *always* care about the specific eraser type (especially in the `has_eraser` case), and these functions cater exactly to those situations. I don't see much sense in requiring callers to re-implement their functionality in order to avoid a compatibility issue which probably doesn't exist.

jigpu: Any feedback, @whot?

libwacom/libwacom-database.c
```diff
@@ -266,25 +266,38 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
-			stylus->has_eraser = FALSE;
-			stylus->has_lens = FALSE;
-			stylus->has_wheel = FALSE;
+		string_list = g_key_file_get_string_list (keyfile, groups[i], "PairedStylusIds", NULL, NULL);
+		if (string_list) {
+			GArray *array;
+			guint i;
+
+			array = g_array_new (FALSE, FALSE, sizeof(int));
+			for (i = 0; string_list[i]; i++) {
```
whot, line 276: `i` is already used for the surrounding `groups` iteration. Which is fine because it'll be shadowed but we shouldn't do this anyway, not least because the `g_warning` below will print the wrong `groups[i]`. (This issue is currently in place with the axis type parsing too, see #248)

libwacom/libwacom-database.c
whot: i noticed you fixed this in the hunk above but not here - any specific reason for that?

libwacom/libwacom-database.c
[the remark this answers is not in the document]
whot: uhm, where? :)

libwacom/libwacom-database.c
jigpu: Fixed

libwacom/libwacom-database.c
```diff
@@ -103,6 +103,20 @@ type_from_str (const char *type)
...
+{
+	if (type == NULL)
+		return WACOM_ERASER_NONE;
+	if (streq(type, "None"))
+		return WACOM_ERASER_NONE;
+	if (streq(type, "Invert"))
+		return WACOM_ERASER_INVERT;
+	if (streq(type, "Button"))
+		return WACOM_ERASER_BUTTON;
+	return WACOM_ERASER_UNKNOWN;
```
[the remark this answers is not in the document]
jigpu: The unknown case is symmetric with other enums (e.g. `WCLASS_UNKNOWN` and `WSTYLUS_UNKNOWN`, neither of which should ever occur). They are used by the test suite to highlight database typos. I do notice, however, that the `WACOM_ERASER_UNKNOWN` case was only implicitly tested for, so I've added an explicit test to cover it. Does that sound okay?

libwacom/libwacom.c
```diff
@@ -1225,7 +1225,7 @@ libwacom_stylus_has_eraser (const WacomStylus *stylus)
 LIBWACOM_EXPORT int
```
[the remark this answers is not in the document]
jigpu: This is next to that hill you're avoiding ;) We could deprecate this and tell callers to instead loop over `get_paired_ids` while checking for `is_eraser`, but I think this still has use as a convenience function. Of course, like `is_eraser` the semantic meaning would slightly change from "has an eraser that you have to invert" to "has an eraser of any description".

jigpu:
Updated branches on top of master and made final changes described above. Also updated "test: Add test to check that erasers have corresponding styli" to have the AES check that it was hoping for, as well as cleaned up commit messages to be ready for merging.

jigpu:
Hopefully final round of changes. Updated #178 to add the AES stylus group to the new sensors we've added to the database, and tacked a documentation patch onto the end of this series to clarify what the library means by "has eraser" and "is" eraser".

Pinglinux:
For EraserType=Button, does it mean a button has been reported as an eraser from the kernel, or does it also mean we can assign a button to an eraser, but the button was not reported as an eraser?

jigpu:
> For EraserType=Button, does it mean a button has been reported as an eraser from the kernel, or does it also mean we can assign a button to an eraser, but the button was not reported as an eraser?

Libwacom describes the hardware, not what software may or may not be capable of. The `EraserType=Button` property only says that the stylus has implemented its eraser with a button. If a particular piece of software enables you to assign some other button to eraser duty, that is not something that libwacom can or should report.

PR #194, opened 2020-01-07
data: Add support for Wacom One Pen Display 13

skomra:
Add .tablet file and add it to meson.build.

Signed-off-by: Aaron Armstrong Skomra <aaron.skomra@wacom.com>


whot:
merging because the CI failure is a meson bug, not an issue with this PR.

PR #195, opened 2020-01-09
ELAN-262B tablet preset

hellcp:
Another one of ELAN devices for Envy x360 13-ag0xxx. I am however facing an issue where I can't see any settings for the stylus (in Gnome Control Center), which are working for my external wacom tablet. I might need some help getting the necessary info :P


whot:
for the actual commit: the addition is there in `meson.build` but not sorted the same way. Diff that complains is this one here:
```diff
-- stdout ---
--- /dev/fd/63	2020-01-28 02:55:29.230858169 +0000
+++ /dev/fd/62	2020-01-28 02:55:29.230858169 +0000
@@ -67,2 +67,3 @@
 data/elan-29b6.tablet
+data/elan-262b.tablet
 data/elan-264c.tablet
@@ -73,3 +74,2 @@
 data/elan-5515.tablet
-data/elan-262b.tablet
 data/generic.tablet
-------
```
Can't remember which direction this diff works but you'll be able to figure it out quickly based on this :)

Other than that the commit looks correct for the libwacom part. Does `libwacom-list-local-devices` list the device after you added the `.tablet` file?

hellcp:
Yup, it shows up just fine, and the stylus now also shows up in gnome control center (and I have no idea what even changed between now and back then).

whot:
Moon phase, I guess. Let's hope the moon stays this way then.

PR #197, opened 2020-01-11
data: Add Intuos Pro 5 M wireless

battlesnake:
Duplicated Intuos Pro M profile to new profile and set ID for when used wirelessly.

### DIFF
--- a/data/intuos-pro-m-wireless.tablet
+++ b/data/intuos-pro-m-wireless.tablet
@@ -1,0 +1,68 @@
+# Wacom
+# Intuos Pro M
+# PTH-651
+#
+# Button Map:
+# (A=1, B=2, C=3, ...)
+#
+#    *-----------------------*
+#    |                       |
+#  B |                       |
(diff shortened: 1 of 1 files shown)


skomra:
0x84 is the PID of the dongle, which I think works with another device as well. I did a quick test on Ubuntu 18.04 and I was able to map the buttons for this device used over wireless. So I think this device is already supported.

whot:
ah, cool, thanks. do we have a list of dongles? would make sense to put in a CI check for the dongle IDs so we prevent any inadvertent merges?

Pinglinux:
As far as I know, there is only wireless dongle, which is 0x84.

whot:
Receiver check was added in #199

whot:
I'm closing this, based on https://github.com/linuxwacom/libwacom/pull/197#issuecomment-573764511 and the fact that this one adds the dongle, not the actual device. Thanks though.

PR #199, opened 2020-01-20
data: add a test for receiver IDs

whot:
We don't want any .tablet entries with the receiver ID as that may match
multiple devices. So let's add a check for it, conveniently integrated via
pytest so all we need is invoke pytest and it'll search recursively for tests
to execute.

The actual test searches the directory for .tablet files and runs each through
a config parser to extract the matches.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ jobs:
   include:
     - stage: test
       install:
-        - sudo apt-get install -y ninja-build python3-pip python3-setuptools valgrind
+        - sudo apt-get install -y ninja-build python3-pip python3-setuptools python3-pytest valgrind
         - sudo pip3 install meson
       script:
         - meson builddir
--- a/data/test_receiver_id.py
+++ b/data/test_receiver_id.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+#
+# Copyright © 2018 Red Hat, Inc.
+#
+# Permission to use, copy, modify, distribute, and sell this software
+# and its documentation for any purpose is hereby granted without
+# fee, provided that the above copyright notice appear in all copies
+# and that both that copyright notice and this permission notice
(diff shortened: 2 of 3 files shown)


jigpu:
Not sure why this isn't just a part of `test/test-tablet-validity.c`, but I guess its fine...

I mocked up an alternative implementation (https://github.com/jigpu/libwacom/commit/d0ae995fbccc325a67e30f66803b4f6744b162e5) and wound up founding that it triggers a bunch of test failures due to WacomMatch objects with non-zero VID:PID despite being on the serial bus (e.g. see `data/isdv4-e3.tablet`)... I see that this is an error for WacomDevice objects, and I'd assume that it should also be an error for WacomMatch objects too but I'm not 100% sure...

Feel free to ignore my implementation if its bogus, however...

whot:
> Not sure why this isn't just a part of test/test-tablet-validity.c, but I guess its fine...

Primarily because I didn't even think of it :) I wrote a similar test for libratbag and immediately thought I could just copy that one and done. Putting it into the validity test works fine for me.

Yeah, I'm pretty sure those nonzero serial vids are a bug and only passed because they are the second match.

jigpu:
After poking around some more I'm actually beginning to wonder if non-zero VID/PID really is a bug... It looks like at one point (a49d218ab2aaba210764e3e8f696d7fb65cfe072) libwacom was able to get this info from udev for inputattach-connected devices and only used 0:0 as a fallback. The data is [still available](https://elixir.bootlin.com/linux/v5.0.1/source/drivers/input/touchscreen/wacom_w8001.c#L550) from the w8001 driver, and the current libwacom code doesn't look like its changed enough to work differently.

I suspect I should update the VID:PID test that I was piggypacking off of to allow VID:PIDs >= 0 rather than strictly == 0.

whot:
closing in favour of #206

PR #206, opened 2020-01-30
Several test suite updates

jigpu:
* Supersedes pull requests #197 and #199 to add a test which checks for the wireless dongle.
* Allow matches to non-zero VID:PID for serial devices as mentioned in #199 
* Expand VID:PID checks to all defined matches

### DIFF
--- a/test/test-tablet-validity.c
+++ b/test/test-tablet-validity.c
@@ -159,23 +159,21 @@ test_name(gconstpointer data)
 }
 
 static void
-test_vidpid(gconstpointer data)
+assert_vidpid(WacomBusType bus, int vid, int pid)
 {
-	WacomDevice *device = (WacomDevice*)data;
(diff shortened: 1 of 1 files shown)


test/test-tablet-validity.c
```diff
@@ -169,6 +169,9 @@ test_vidpid(gconstpointer data)
 			g_assert_cmpint(libwacom_get_product_id(device), >=, 0);
 			break;
 		case WBUSTYPE_USB:
+			if (libwacom_get_vendor_id(device) == 0x056A)
+				g_assert_cmpint(libwacom_get_product_id(device), !=, 0x84);
+			/* fall through */
```
whot: Please put a comment here that `0x84` is the wireless dongle so others don't have to go through `git blame` to figure that out.

PR #216, opened 2020-02-24
Unbreak build with GCC < 5

jbeich:
GCC < 5 uses `-std=gnu89` by default while GCC >= 5 uses `-std=gnu11` by default.
https://gcc.gnu.org/gcc-5/changes.html


configure.ac
```diff
@@ -22,6 +22,8 @@ AC_PROG_CC
 AC_PROG_INSTALL
 PKG_PROG_PKG_CONFIG
 
+CFLAGS="$CFLAGS -std=gnu99"
+
```
whot: quick google suggests we should be using `AC_PROG_CC_C99` here, right? or rather `AC_PROG_CC_STDC` though that's less specific.

jbeich: `AC_PROG_CC_STDC` doesn't pass `-std=gnu99` if compiler defaults to `-std=gnu11` which may lead to a regression due to discrepancy with Meson build.

jbeich:
Oops, accidentally pushed unrelated commit. I'll removed it.

whot:
Thanks!

PR #219, opened 2020-03-02
Unbreak Meson build if libxml2 is not installed

jbeich:
[Found](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=244554) downstream. libxml2 dependency has been there for ages but Meson made it mandatory.


whot:
thanks!

PR #220, opened 2020-03-05
Unbreak autotools build on BSDs

jbeich:
See find(1) manpage on [DragonFly](https://www.dragonflybsd.org/cgi/web-man?command=find&section=1), [FreeBSD](https://man.freebsd.org/find/1), [NetBSD](https://netbsd.gw.com/cgi-bin/man-cgi?find+1+NetBSD-current) [OpenBSD](https://man.openbsd.org/find.1). While build doesn't actually fail despite errors some files end up not being installed e.g., https://github.com/freebsd/freebsd-ports/commit/eea904e27f1b

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -1,9 +1,9 @@
 libwacomtabletdir = $(datadir)/libwacom
-tablet_files = $(shell find $(top_srcdir)/data -name "*.tablet" -printf "%P\n")
+tablet_files = $(shell cd $(top_srcdir)/data; printf '%s\n' *.tablet)
 dist_libwacomtablet_DATA =  $(tablet_files)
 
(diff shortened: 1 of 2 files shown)


data/Makefile.am
```diff
@@ -1,9 +1,9 @@
 libwacomtabletdir = $(datadir)/libwacom
-tablet_files = $(shell find $(top_srcdir)/data -name "*.tablet" -printf "%P\n")
+tablet_files = $(shell cd $(top_srcdir)/data; printf '%s\n' *.tablet)
```
jbeich: Ideally, `cd foo; printf '%s\n' *.ext` could be replaced by `basename foo/*.ext` but GNU version of `basename` doesn't support more than one argument.

whot:
thanks!

PR #229, opened 2020-03-24
libwacom 1.3

whot:


whot:
opened this to run travis, but that doesn't seem to happen...

and since github always rebases and changes the sha my tag would be lost, so I'm going to push this directly.

PR #230, opened 2020-03-26
data: Add elan-0732 (HP ENVY x360 Convertible 13-ag0xxx)

ColinKinloch:
Seems to work on my envy x360, I just copied elan-262b.tablet and changed the device match.
Is there anything else I need to check?

### DIFF
--- a/data/elan-0732.tablet
+++ b/data/elan-0732.tablet
@@ -1,0 +1,15 @@
+# ELAN touchscreen/pen sensor present in the HP envy x360 Convertible 13-ag0xxx
+[Device]
+Name=ELAN 0732
+ModelName=
+DeviceMatch=i2c:04f3:271c
+Class=ISDV4
+Width=12
+Height=7
+IntegratedIn=Display;System
+
+[Features]
+Stylus=true
+Touch=true
+Buttons=0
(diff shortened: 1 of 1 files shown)


whot:
unrelated to the patch here: @jigpu are you seeing travis results in PRs? this is the second one where I noticed nothing shows up.

whot:
fwiw, I tested this locally and the script disagreed about the sorting order. Which caused me to look up meson things again and eventually resulted in #232. Once that's merged you can just drop the second commit here and  we can merge this one.

whot:
rebased locally and test suite succeed, merging. Thanks!

PiFern:
Hi ! I'm not sure if it's the right place to post, please tell me if it's not...
On my computer "HP Pavilion x360 Convertible", tablet and stylus are working (even if the stylus isn't really precise) using the "serial-wacf004.tablet" file, but "xinput --list" returns "ELAN0732:00 04F3:251C" for the device, so "04f3:251c" should be added to "elan-0732.tablet" file ? Is there anything else to do to make my tablet well handled ?
"xsetwacom --list" doesn't return anything, i don't know if it should ...

PR #232, opened 2020-03-27
use meson's install_subdir instead of naming all tablet files

whot:
I think our life just got a little bit better :)

This is on top of #231 because I don't want to have to fix the conflicts.

It'll also now cause a conflict with #230, sorry about that @ColinKinloch but you can just drop the second patch.


jigpu:
LGTM

PR #233, opened 2020-03-31
tools: switch from a giant udev rulest to using hwdb entries

whot:
More efficient and easier to understand and read too. Plus, this set works
with uinput devices out of the box without the need for extra rules.

The new approach is split into a ruleset which now only tags the old serial
devices. Those two lines are easier to keep here than device a hwdb lookup
scheme.

The ruleset imports from the hwdb builtin with the device's modalias, see
/sys/class/device/input/eventXXX/device/modalias. This provides the
bus/vendor/product lookup. And for the prefix use libwacom with the device
name which allows our Finger and Pad matches.

Since we cannot unset properties through a hwdb lookup, we have an extra rule
to set all the joystick 0 envs to the empty string.

The hwdb file itself is generated, note that you must to run `systemd-hwdb`
update whenever this file changes.


jigpu:
Will this affect users of distros and OSes which don't use systemd?

whot:
it shouldn't, those still use udev and the hwdb is a pretty integral part (all laptop key overrides for crazy keys are in the hwdb).

jigpu:
Had a play around with hwdb on a Devuan system to get more familiar with it. After an embarrassingly long time flailing around, I'm happy enough to give my Acked-by :+1:

whot:
thanks!

on that note - I wonder if it's worthy keeping the generated hwdb file in git so on the target machine we only have to install it and we know exactly what it contains at any stage?

PR #235, opened 2020-04-16
Huion hs610

p-velasco:
I wrote the .svg and .tablet files for the Huion HS610. However I'm not completely sure that the CW and CCW buttons works, as right now my gnome 3.30 don't show them on the config panel and the mode switching feature doesn't work. But all the buttons can be assigned and the layout is correct.

### DIFF
--- a/data/huion-hs610.tablet
+++ b/data/huion-hs610.tablet
@@ -1,0 +1,26 @@
+# HUION
+# HS610
+#
+
+[Device]
+Name=Huion HS610
+DeviceMatch=usb:256c:006d:HUION Huion Tablet Pad;usb:256c:006d:HUION Huion Tablet Pen
+Class=Bamboo
+Width=10
+Height=6
+IntegratedIn=
+Layout=huion-hs610.svg
+Styli=0xfffff
+
+[Features]
+Stylus=true
+Reversible=true
+Touch=false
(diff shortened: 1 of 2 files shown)


javierpena:
It's interesting. I just purchased a Huion HS64 pad and was looking at adding support for it, and it looks like it uses the same USB ID (256c:006d). However, it has a different size (6x4 inches) and number of buttons (4 vs 13). How could we distinguish between the two?

whot:
huion reuses the usb vid/pid on most (all?) tablets, we have to distinguish those by name which works on most devices. see the existing `.tablet` file for huions, same should be possible for yours I hope.

javierpena:
I'm afraid we've been unlucky this time, the name it uses is "HUION Huion Tablet Pen" and "HUION Huion Tablet Pad", just like this model.

whot:
whoop-dee-doo...

fwiw, size doesn't really matter since no-one uses the libwacom entry here afaik. buttons do matter but we'll have to err on the side of the device with more buttons to make sure they all show up in the config interfaces.

whot:
Closing, it's been 2 years since the last comment and we've merged a bunch of HUION tablets since. Might be that this one was part of it, if not please re-open or open a new PR so we can start the discusions anew.

PR #236, opened 2020-04-17
data: Add Huion 420

Xyene:
I've omitted a layout SVG, since this tablet has no buttons or anything but the active area. Hopefully that's alright, if not I can make one. `Styli` was copied over from the H420 definition.


whot:
thanks!

PR #237, opened 2020-04-17
data: fix typo in Huion H420 SVG title

Xyene:
Noticed this minor typo while looking for layout examples. As far as I can find, the H410 is not an existing device, and most likely a typo.


whot:
thanks!

PR #239, opened 2020-04-19
tools: expand Huion consumer control quirk to include GAOMON

Xyene:
For instance, a GAOMON S56K device exposes "Tablet Monitor Consumer
Control" and "Tablet Monitor System Control" devices, both of which are
not tablet devices.

Instead of listing the Cartesian product of [HUION, TabletMonitor] &times; [Consumer Control, System Control] (as far as I can tell the udev syntax doesn't allow regex-like groupings?), I relaxed the matching to exclude all Consumer/System Control devices. I can write the Cartesian product directly if that's preferred, though.

### DIFF
--- a/tools/65-libwacom.rules.in
+++ b/tools/65-libwacom.rules.in
@@ -3,8 +3,9 @@
 ACTION!="add|change", GOTO="libwacom_end"
 KERNEL!="event[0-9]*", GOTO="libwacom_end"
 
-# HUION consumer control devices are not tablets.
-ATTRS{name}=="HUION * Consumer Control", GOTO="libwacom_end"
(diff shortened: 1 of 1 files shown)


Xyene:
Done :+1:

whot:
> Possibly related, but the Huion 420/GAOMON S56K says it has a mouse device that doesn't actually exist, and makes libinput give (harmless enough) warnings:

HUION devices are... inconsistent. What happens there is likely that it exposes some event nodes that cause udev to label it `ID_INPUT_TABLET` but it doesn't actually have tablet capabilities otherwise. Usually you can ignore that warning, it's not worth the effort of fixing

PR #241, opened 2020-04-20
Build system fixes

jigpu:
Fixes a warning from Meson when configuring the project, and corrects build issues that have just popped up with Travis


whot:
Thanks!

ftr, there's an option for `pip` to install a specific version only, but in this case it's better to stay up-to-date to catch errors early.

PR #243, opened 2020-04-25
Support Wacom Volito 4x5 (FT-0405, 056a:0060)

mkszuba:
A pretty ancient and simple (it has a pen with no eraser and only one button) device but in the pre-libinput (or perhaps legacy-driver) days it worked perfectly - I was able to, out of the box, draw with varying pressure in the GIMP and the on-stylus button correctly worked as the right-mouse button. Getting it to work on a modern system took some work (in case anyone is interested, turns out that in order to both have the driver trigger button-1 events on contact with the tablet surface rather than as soon as the stylus came into proximity, and to get any pressure readings other than 1 once down, I had to change the threshold from the default 26 to 5 _exactly_ - no other value I have tried seems to have any effect) but in the end it does work too, and as a bonus it seems the generic no-eraser stylus definition suffices in spite of the Volito stylus (ID 0x2, in case anyone wants to know) only having one button - we simply never get any events from the second button.

### DIFF
--- a/data/volito-4x5.tablet
+++ b/data/volito-4x5.tablet
@@ -1,0 +1,15 @@
+# Wacom
+# Volito 4x5
+# FT-0405
+
+[Device]
+Name=Wacom Volito 4x5
+ModelName=FT-0405
+DeviceMatch=usb:056a:0060
+Width=5
+Height=4
+Styli=0xffffd;
+
+[Features]
+Stylus=true
+Buttons=0


mkszuba:
Turns out that the stylus button works fine out of the box even using the generic stylus entry, the problem was simply that the GIMP (or more precisely, X) got confused by having both LMB (from proximity) and RMB (from the button itself) pressed at the same time. Once I disabled the former using _xsetwacom_, right-clicks began to work fine... except with tip-down events generating no button presses, one cannot actually draw anything.

I do not know what is going on here.

jigpu:
Devices like the Volito, Graphire, and Bamboo use pens that have no intrinsic identifiers. Furthermore, the tablets themselves may support both 1-button and 2-button pens, but since there's no way for libwacom or the driver to distinguish them, the stylus definition file only includes 2-button variants just to be sure userspace never gets surprised by a button it thinks shouldn't exist.

The tablet definition looks reasonable, but I'll let @whot have a second look before merging.

As for the issues you're having with the buttons (needing to change the default threshold), please open an issue at the [input-wacom bug tracker](https://github.com/linuxwacom/input-wacom/issues/) with a recording from [capture.sh](https://gist.githubusercontent.com/jigpu/b7b47e4bede584cec8562f666fd84af4/raw/af36c0a0fc2c019750472443369915ad7ea97737/capture.sh) of you bringing the pen into and out of prox without actually touching it to the tablet.

whot:
LGTM too, thanks

PR #248, opened 2020-05-15
Fix a test suite failure and a re-used loop variable

whot:


data/volito-4x5.tablet
```diff
@@ -6,6 +6,7 @@
 Name=Wacom Volito 4x5
 ModelName=FT-0405
 DeviceMatch=usb:056a:0060
+Class=Bamboo
```
jigpu, line 9: Probably better to be marked as "Graphire" class. Otherwise, LGTM
Reviewed-by: Jason Gerecke <jason.gerecke@wacom.com>

whot: done, thanks

PR #250, opened 2020-05-16
Add HP ZBook Studio x360 G5 Wacom AES sensor 48d6

Eloston:
Add AES sensor used by the HP ZBook Studio x360 G5.

Tested with the HP Rechargeable Active Pen G3 (0x826B). Aside: I noticed that this stylus is listed in #178 without a name; is there some info I could provide to document it better?

For reference, I've attached [record_1589605219.tar.gz](https://github.com/linuxwacom/libwacom/files/4637982/record_1589605219.tar.gz) produced by [the caputure.sh script](https://gist.githubusercontent.com/jigpu/b7b47e4bede584cec8562f666fd84af4/raw/af36c0a0fc2c019750472443369915ad7ea97737/capture.sh).

### DIFF
--- a/data/isdv4-48d6.tablet
+++ b/data/isdv4-48d6.tablet
@@ -1,0 +1,16 @@
+# Active electrostatic (AES) sensor used by the HP ZBook Studio x360 G5
+
+[Device]
+Name=Wacom ISDv4 48d6
+ModelName=
+DeviceMatch=i2c:056a:48d6
+Class=ISDV4
+Width=14
+Height=8
+IntegratedIn=Display;System
+Styli=@isdv4-aes;
+
+[Features]
+Stylus=true
+Touch=true
+Buttons=0


whot:
can you please rebase your branch on top of master to re-trigger the CI. we had some issues with that, fixed now.

jigpu:
Looks good to me, though the patch will need to be updated to include a `Styli=@isdv4-aes;` line just below the `IntegratedIn` line now that #178 has been merged. Aside from that, it looks good to me :+1: 

As for the stylus, can you double-check the name and ID? It looks like 0x846B is also supposed to be an "HP Rechargeable Active Pen G3". They could have made multiple styli with that name, but doesn't hurt to check. Also, if you can find a model number on/inside the pen (e.g. GSG43AA is the 0x846B) that would help to distinguish it as well.

Eloston:
@jigpu The pen is an optional add-on with the ZBook Studio x360 G5. On the packing list for the laptop, it has part number "2YS89AV" (which has a number of results on Google) with name "ZBook Studio x360 Pen".

However, the sticker on the pen itself says it is an "HP ACTIVE PEN G3", which is consistent with the Bluetooth device name that it advertises.

To the right of the sticker are three codes printed directly on the pen itself, two of which are "L57042-001" and "L08263-003" (which have some results on Google).

Let me know if you need more info.

jigpu:
Thanks @Eloston. I've opened a pull request at #270 which updates the comments for 0x826B.

PR #253, opened 2020-05-17
Add FreeBSD to CI

jbeich:
TravisCI [supports](https://github.com/travis-ci/travis-build/pulls?q=is%3Apr+freebsd+is%3Aclosed) FreeBSD nowadays. Let's see how well it works.

CC @zeising

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,6 @@ compiler:
   - gcc
   - clang
 
-addons:
-  apt:
-    packages:
-      - libgudev-1.0-dev
-      - libxml++2.6-dev
-
 stages:
   - test
   - name: coverity
@@ -26,6 +20,11 @@ env:
     - CFLAGS="-Werror -Wall -Wextra -Wno-error=sign-compare -Wno-error=unused-parameter -Wno-error=missing-field-initializers"
 
 before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" == linux ]; then
+      sudo apt-get install -yq --no-install-suggests --no-install-recommends \
+        libgudev-1.0-dev libxml++2.6-dev
+    fi
   - |
     if [ "$TRAVIS_OS_NAME" == freebsd ]; then
       sudo pkg install -y evdev-proto libgudev libxml++
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ sudo: required
 dist: xenial
 language: c
 
+os:
+  - linux
+  - freebsd
+
 compiler:
   - gcc
   - clang
@@ -21,6 +25,13 @@ env:
   global:
     - CFLAGS="-Werror -Wall -Wextra -Wno-error=sign-compare -Wno-error=unused-parameter -Wno-error=missing-field-initializers"
 
+before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" == freebsd ]; then
+      sudo pkg install -y evdev-proto libgudev libxml++
+      sudo ln -sf $(which gmake) $(which make)
+    fi
+
 script:
   - mkdir _build
   - pushd _build > /dev/null
@@ -32,6 +43,16 @@ script:
 
 jobs:
   include:
+    # only test build using meson on freebsd, the rest is covered by linux
+    - stage: test
+      os: freebsd
+      compiler: clang
+      install:
+        - sudo pkg install -y meson
+      script:
+        - meson builddir || (cat builddir/meson-logs/meson-log.txt && false)
+        - ninja -C builddir test || (cat builddir/meson-logs/testlog.txt && false)
+        - ninja -C builddir dist
     - stage: test
       install:
         - sudo apt-get install -y python3-pip python3-setuptools valgrind


.travis.yml
```diff
@@ -11,6 +15,11 @@ addons:
     packages:
       - libgudev-1.0-dev
       - libxml++2.6-dev
+  pkg:
+    packages:
+      - evdev-proto
```
jbeich, line 20: Required since f9cf43e43171 due to missing include/linux/input.h.

jbeich:
Hmm, `update: true` didn't help FreeBSD jobs. `Installing 3 packages` step doesn't show up in TrivisCI logs which leads to bustage due to missing deps:
```
checking for GLIB... no
configure: error: Package requirements (glib-2.0 gudev-1.0) were not met:
Package 'glib-2.0', required by 'virtual:world', not found
Package 'gudev-1.0', required by 'virtual:world', not found
```
I can't reproduce in my fork.

jbeich:
All green now if a bit ugly (manual install, `make` symlink).

.travis.yml
```diff
@@ -44,6 +44,16 @@ script:
 
 jobs:
   include:
+    # only test build using meson on freebsd, the rest is covered by linux
+    - stage: test
+      os: freebsd
+      compiler: clang
+      install:
+        - sudo pkg install -y meson
```
whot, line 52: we install from pip for ubuntu (because it's old), for consistency we should do the same here I think

.travis.yml
```diff
@@ -21,6 +25,14 @@ env:
   global:
     - CFLAGS="-Werror -Wall -Wextra -Wno-error=sign-compare -Wno-error=unused-parameter -Wno-error=missing-field-initializers"
 
+before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" == freebsd ]; then
+      sudo pkg install -y evdev-proto libgudev libxml++
+      sudo pkg install -y automake libtool pkgconf gmake
+      sudo ln -sf $(which gmake) $(which make)
```
whot, line 33: I forgot: aren't we root anyway? do we even need sudo here? it's used elsewhere already but might be unnecessary there too.

.travis.yml
jbeich: FreeBSD doesn't support old packages: `/quarterly` (default) has meson 0.53.2, `/latest` has 0.54.2. Installing from pip means trusting upstream projects to QA on FreeBSD but Meson project doesn't seem to.

.travis.yml
jbeich: I've tried: `pkg install` failed without `sudo`. [Official examples](https://docs.travis-ci.com/user/installing-dependencies/) also use `sudo`.

jbeich:
> if `addons` doesn't handle freebsd correctly

Or maybe `pkg` addon haven't propgated to all TravisCI builders yet. CirrusCI has less bugs but unless linux is also converted it's not going to decrease maintenance.

> move things out in a more generic manner

Pointless abstraction. ;) When s/apt-get/pkg/ is not enough it maybe worth to move logic into separate scripts. Such scripts would be easier to reuse with more than one CI.

.travis.yml
```diff
@@ -21,6 +25,14 @@ env:
   global:
     - CFLAGS="-Werror -Wall -Wextra -Wno-error=sign-compare -Wno-error=unused-parameter -Wno-error=missing-field-initializers"
 
+before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" == freebsd ]; then
+      sudo pkg install -y evdev-proto libgudev libxml++
+      sudo pkg install -y automake libtool pkgconf gmake
```
jbeich, line 32: TravisCI have these dependencies preinstalled, so I can drop it to avoid confusion. For example, `bash` used to run libwacom test scripts is actually what TravisCI use to run `before_install` target.

.travis.yml
whot: oki, thanks for testing

.travis.yml
whot: ack

jbeich: Language-specific package managers tend to have trouble with foreign build systems on non-mainstream platforms e.g.,
```
$ sudo pkg install -qy py37-pip
$ pip install meson ninja
Collecting meson
  Downloading https://files.pythonhosted.org/packages/78/14/53d77eeb196efb1230eedaba556c920fc59989f1ebd429e990cb129b41d3/meson-0.54.2.tar.gz (1.7MB)
     |████████████████████████████████| 1.7MB 724kB/s
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Collecting ninja
  Downloading https://files.pythonhosted.org/packages/dd/96/e2ec4acccb8dee33b4987f553d531d61e3081c8d4cfbce249655dfe23906/ninja-1.9.0.post1.tar.gz
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-ef8x46bl/ninja/setup.py", line 7, in <module>
        from skbuild import setup
    ModuleNotFoundError: No module named 'skbuild'
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-ef8x46bl/ninja/
```

.travis.yml
```diff
@@ -26,6 +20,10 @@ env:
     - CFLAGS="-Werror -Wall -Wextra -Wno-error=sign-compare -Wno-error=unused-parameter -Wno-error=missing-field-initializers"
 
 before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" == linux ]; then
+      sudo apt-get install -y libgudev-1.0-dev libxml++2.6-dev
```
jigpu, line 25: Any particular reason for using the `-y` option here (and in the FreeBSD block)? I'd like to keep the number of pulled in packages to a minimum just to be sure we don't introduce any accidental build dependencies.

.travis.yml
```diff
@@ -29,7 +29,6 @@ before_install:
   - |
     if [ "$TRAVIS_OS_NAME" == freebsd ]; then
       sudo pkg install -y evdev-proto libgudev libxml++
-      sudo pkg install -y automake libtool pkgconf gmake
```
jigpu: Can we fixup the first commit with this change instead of introducing it separately?

.travis.yml
jbeich: To avoid the job timing out on user input. `apt` addon also uses `-y`.

.travis.yml
jbeich: For easier re-review. I've squashed all FreeBSD commits.

.travis.yml
jbeich: Also, I'm not very familar with `apt-get`. Maybe it assumes `-y` if run non-interactively but the manpage doesn't explicitly say so.

jbeich: I've added `-q` and `--no-install-suggests --no-install-recommends` as `apt` addon did.

jigpu:
Thanks, merged.

PR #256, opened 2020-06-04
Fix the version number for the new API

whot:
The version number is technically meaningless since it's just a string. But we
try to keep it in sync with the version the API was introduced in.


whot:
Ironically enough, this is an ABI break now but we don't need to worry about those on the master branch between releases.

PR #257, opened 2020-06-06
Add Huion New 1060 Plus

real-or-random:
This adds the Huion New 1060 Plus. I've tested this with the equivalent Gaomon M106K but I've added this as Huion as these seem to be more common. The layout for the Gaomon is be very similar: it has round buttons instead of rectangular buttons but they're in the same locations, so I don't think it's worth the effort to create a different layout file (and I'm not willing to spend any more time on tuning SVGs ;)).

There are a few things where I'm not 100% sure, I'll comment inline.

### DIFF
--- a/data/huion-new-1060-plus.tablet
+++ b/data/huion-new-1060-plus.tablet
@@ -1,0 +1,24 @@
+# HUION
+# New 1060 Plus
+#
+
+[Device]
+Name=Huion New 1060 Plus
+ModelName=
+DeviceMatch=usb:256c:006e:HID 256c:006e Pen;usb:256c:006e:HID 256c:006e Pad
+Class=Bamboo
+Width=10
+Height=6
+IntegratedIn=
+Layout=huion-new-1060-plus.svg
+Styli=0xffffd;
+
+[Features]
+Stylus=true
+Reversible=true
+Touch=false
+Buttons=12
+
+[Buttons]
+Left=A;B;C;D;E;F;G;H;I;J;K;L
+EvdevCodes=0x100;0x101;0x102;0x103;0x104;0x105;0x106;0x107;0x108;0x109;0x130;0x131
--- a/data/layouts/huion-new-1060-plus.svg
+++ b/data/layouts/huion-new-1060-plus.svg
@@ -1,0 +1,180 @@
+<?xml version="1.0" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
+   "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg
+   width="360.0"
+   height="240.0"
+   style="color:#000000;font-size:8px;fill:none;stroke:#7f7f7f;stroke-width:0.25"
+   version="1.1"
+   xmlns="http://www.w3.org/2000/svg">
+  <title
+     id="title">Huion New 1060 Plus</title>
+  <path
+     id="ButtonA"
+     class="A Button"
+     d="M 16.0 42.5 H 32.0 v 12.0 H 16.0 c -0.5 0.0 -1.0 -0.5 -1.0 -1.0 v -10.0 c 0.0 -0.5 0.5 -1.0 1.0 -1.0 z" />
+  <path
+     id="LeaderA"
+     class="A Leader"
+     d="M 8.5 51.5 H 13.5" />
+  <text
+     id="LabelA"
+     class="A Label"
+     x="7.0"
+     y="51.5"
+     style="text-anchor:end">A</text>
+  <path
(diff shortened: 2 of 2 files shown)


data/huion-new-1060-plus.tablet
```diff
@@ -0,0 +1,23 @@
+# HUION
+# New 1060 Plus
+#
+
+[Device]
+Name=Huion New 1060 Plus
+ModelName=
+DeviceMatch=usb:256c:006e:HID 256c:006e Pen;usb:256c:006e:HID 256c:006e Pad
+Class=Bamboo
```
real-or-random, line 9: This works but I have no idea what the right `Class` is. I can't find docs on this.

data/huion-new-1060-plus.tablet
```diff
@@ -0,0 +1,23 @@
+# HUION
+# New 1060 Plus
+#
+
+[Device]
+Name=Huion New 1060 Plus
+ModelName=
```
real-or-random, line 7: Is it okay to leave this empty? What should go here?

data/layouts/huion-new-1060-plus.svg
```diff
@@ -0,0 +1,180 @@
...
+  <path
+     id="LeaderB"
+     class="B Leader"
+     d="M 56.5 51.5 H 51.5" />
+  <text
+     id="LabelB"
+     class="B Label"
+     x="57.0"
+     y="51.5"
+     style="text-anchor:start">B</text>
```
real-or-random, line 39: I think `text-anchor:start` is the right thing for the right buttons but the gnome-settings dialog puts the text to the left of this such that it overlays the button path. Am I wrong here or is this is a bug in gnome?

data/layouts/huion-new-1060-plus.svg
```diff
@@ -0,0 +1,180 @@
...
+     id="ButtonA"
+     class="A Button"
+     d="M 16.0 42.5 H 32.0 v 12.0 H 16.0 c -0.5 0.0 -1.0 -0.5 -1.0 -1.0 v -10.0 c 0.0 -0.5 0.5 -1.0 1.0 -1.0 z" />
+  <path
+     id="LeaderA"
+     class="A Leader"
+     d="M 8.5 51.5 H 13.5" />
+  <text
+     id="LabelA"
+     class="A Label"
```
real-or-random, line 22: The README says that I should add `ModeSwitch` if this is a ModeSwitch button. But I have no idea what this is.

data/huion-new-1060-plus.tablet
```diff
@@ -0,0 +1,23 @@
...
+DeviceMatch=usb:256c:006e:HID 256c:006e Pen;usb:256c:006e:HID 256c:006e Pad
+Class=Bamboo
+Width=10
+Height=6
+IntegratedIn=
+Layout=huion-new-1060-plus.svg
+
+[Features]
+Stylus=true
```
real-or-random, line 16: Is it okay not to specify any stylii? It comes with a simple stylus with two buttons and no eraser. gnome shows my the two buttons but it says "General Pen Eraser".

data/huion-new-1060-plus.tablet
whot: nothing uses the `Class` anyway, we just have to keep it in case *something* at some point used it and we don't want to break compatibility. We used `Bamboo` for the other Huions, good enough for me.

data/huion-new-1060-plus.tablet
whot: if the device has a technical model name it should go there. All wacom devices have marketing name "Intuos Pro blah blah" and a model name in the specs "PTK-1234". If the HUION doesn't have one, leave it out.

data/huion-new-1060-plus.tablet
whot: IIRC the huions don't support serial numbers (check for `MSC_SERIAL` in the `libinput record` output of the device). So in that case - use the generic 0xffffd stylus please - see the existing huion file(s) for reference.

data/layouts/huion-new-1060-plus.svg
whot: The mode switch button changes the mode of the ring/strips (if any) and toggles the LEDs arround too. So you can have the ring scroll up/down in mode 1 but zoom in/out in mode 2. Doesn't apply here, so it's not a mode switch.

data/layouts/huion-new-1060-plus.svg
whot: hmm, I think so but I'm not sure myself. firefox and chrome at least appear to render this correctly, so this may be a bug in librsvg. I'd stick with this until proven otherwise ;)

Issue #275, opened 2020-07-04
tablet definition for HP Pavilion x360 14-dh0xxx 


fs3o5:
**Device name**
HP Pavilion x360 14-dh0xxx 

**Device model identifier**
ELAN2514

**udevadm info output**
P: /devices/pci0000:00/0000:00:15.1/i2c_designware.1/i2c-6/i2c-ELAN2514:00/0018:04F3:23DB.0006/input/input29/event9
N: input/event9
S: input/by-path/pci-0000:00:15.1-platform-i2c_designware.1-event
E: DEVLINKS=/dev/input/by-path/pci-0000:00:15.1-platform-i2c_designware.1-event
E: DEVNAME=/dev/input/event9
E: DEVPATH=/devices/pci0000:00/0000:00:15.1/i2c_designware.1/i2c-6/i2c-ELAN2514:00/0018:04F3:23DB.0006/input/input29/event9
E: ID_INPUT=1
E: ID_INPUT_HEIGHT_MM=172
E: ID_INPUT_TOUCHSCREEN=1
E: ID_INPUT_WIDTH_MM=300
E: ID_PATH=pci-0000:00:15.1-platform-i2c_designware.1
E: ID_PATH_TAG=pci-0000_00_15_1-platform-i2c_designware_1
E: LIBINPUT_DEVICE_GROUP=18/4f3/23db:i2c-ELAN2514:00
E: MAJOR=13
E: MINOR=73
E: SUBSYSTEM=input
E: USEC_INITIALIZED=5167157

https://github.com/linuxwacom/wacom-hid-descriptors/issues/85
There is already a tablet definition for the same ELAN2514 but it uses another DeviceMatch.

whot:
Please submit a PR with the extra `DeviceMatch` added, thanks. There are quite a few data files that have multiple matches, pick any of those as an example and check the git history for the common commit formats. Thanks.

PR #286, opened 2020-08-18
Add HP Pavilion x360 Convertible 14-dh0xxx

fs3o5:
https://github.com/linuxwacom/libwacom/issues/275
I wasn't sure what to name the file. Let me know if I should change something.


whot:
Looks good, thanks. The only thing I'd request here: please add the link to the [sysinfo](https://github.com/linuxwacom/wacom-hid-descriptors), we're collecting those so that if anything goes wrong or we need more data in the future, we can emulate the device and extract the required information. See commit 16dd6825a for an example commit message and the comment link. Thanks

whot:
thanks!

Issue #267, opened 2020-06-24
1.4: libwacom still is not LTO ready


kloczek:
It was alredy reported in #152.
```
[tkloczko@barrel libwacom-1.4]$ /usr/bin/ninja -v -j1 -C x86_64-redhat-linux-gnu
ninja: Entering directory `x86_64-redhat-linux-gnu'
[1/2] /usr/bin/meson --internal exe --capture 65-libwacom.hwdb -- /home/tkloczko/rpmbuild/BUILD/libwacom-1.4/x86_64-redhat-linux-gnu/generate-hwdb
[2/2] gcc  -o test-deprecated 'test-deprecated@exe/test_test-deprecated.c.o' -Wl,--as-needed -Wl,--no-undefined -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -flto=auto -flto-partition=none -Os -Wl,-z,relro -Wl,--as-needed -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -flto=auto -flto-partition=none -fuse-linker-plugin -Wl,--start-group libwacom.so.2.6.1 -ldl -Wl,--end-group '-Wl,-rpath,$ORIGIN/' -Wl,-rpath-link,/home/tkloczko/rpmbuild/BUILD/libwacom-1.4/x86_64-redhat-linux-gnu/
FAILED: test-deprecated
[...]

Issue #192, opened 2019-12-25
Test suite fails on musl libc: symbols not found


awilfox:
~~~
FAIL: test-deprecated
=====================

Error relocating ./test-deprecated: libwacom_match_destroy: symbol not found
Error relocating ./test-deprecated: libwacom_update_match: symbol not found
Error relocating ./test-deprecated: libwacom_stylus_destroy: symbol not found
Error relocating ./test-deprecated: libwacom_error_set: symbol not found
Error relocating ./test-deprecated: libwacom_match_new: symbol not found
FAIL test-deprecated (exit status: 127)
~~~

This is caused by:

```c
/**
 * The normal symbols are hidden now when linking against a new library.
 * We know which version they're in though so we can remap them for this
 * test - if we get an unresolved symbol something has gone wrong.
 */
asm(".symver libwacom_match_destroy,libwacom_match_destroy@LIBWACOM_0.33");
asm(".symver libwacom_match_new,libwacom_match_new@LIBWACOM_0.33");
asm(".symver libwacom_error_set,libwacom_error_set@LIBWACOM_0.33");
asm(".symver libwacom_update_match,libwacom_update_match@LIBWACOM_0.33");
asm(".symver libwacom_stylus_destroy,libwacom_stylus_destroy@LIBWACOM_0.33");
```

musl does not support symbol versions like this, so this test necessarily fails when using musl.

whot:
need to disable this test on musl then - any chance you can send us a patch for this?

Issue #170, opened 2019-10-27
fails with gold linker


okias:
issue visible with autotools and meson, both libwacom 1.0 & 1.1.

Linking fails with new gold linker, but works reliably with old BFD.
```
meson --buildtype plain --libdir lib64 --localstatedir /var/lib --prefix /usr --sysconfdir /etc --wrap-mode nodownload /var/tmp/portage/dev-libs/libwacom-1.1/work/libwacom-1.1 /var/tmp/portage/dev-libs/libwacom-1.1/work/libwacom-1.1-build
The Meson build system
Version: 0.52.0
Source dir: /var/tmp/portage/dev-libs/libwacom-1.1/work/libwacom-1.1
Build dir: /var/tmp/portage/dev-libs/libwacom-1.1/work/libwacom-1.1-build
Build type: native build
Project name: libwacom
Project version: 1.1
Appending CFLAGS from environment: '-march=native -O2 -flto -fno-stack-protector -pipe -fno-lto -fno-use-linker-plugin'
Appending LDFLAGS from environment: ['-Wl,-O1', '-Wl,--as-needed', '-fno-lto', '-fno-use-linker-plugin']
Appending CFLAGS from environment: '-march=native -O2 -flto -fno-stack-protector -pipe -fno-lto -fno-use-linker-plugin'
Appending LDFLAGS from environment: ['-Wl,-O1', '-Wl,--as-needed', '-fno-lto', '-fno-use-linker-plugin']
C compiler for the host machine: x86_64-pc-linux-gnu-gcc (gcc 9.2.0 "x86_64-pc-linux-gnu-gcc (Gentoo 9.2.0 p1) 9.2.0")
C linker for the host machine: GNU ld.gold 2.32
Host machine cpu family: x86_64
Host machine cpu: x86_64
Found pkg-config: /usr/bin/x86_64-pc-linux-gnu-pkg-config (0.29.2)
Run-time dependency gudev-1.0 found: YES 233
Run-time dependency glib-2.0 found: YES 2.61.2
Run-time dependency librsvg-2.0 found: YES 2.45.92
Run-time dependency gtk+-2.0 found: YES 2.24.32
Program data/check-files-in-git.sh found: YES (/var/tmp/portage/dev-libs/libwacom-1.1/work/libwacom-1.1/data/check-files-in-git.sh)
[...]

PR #290, opened 2020-09-15
Bump the soname to get rid of the deprecated symbols

whot:
This is an ABI break but doesn't have any effect on any caller, just rebuild
the caller and done. None of these symbols were ever used by anyone (see
d918631fe65594b62283980beef5a14e7f744b97), keeping them was nice for ABI
stability but it also causes some issues, notably with compilers not
supporting symbol versioning.

The new exported symbol set is already named libwacom 2.0. Still the same API
but it makes for less surprises of those using the API.

Fixes https://github.com/linuxwacom/libwacom/issues/192
Fixes https://github.com/linuxwacom/libwacom/issues/170
Fixes https://github.com/linuxwacom/libwacom/issues/267


whot:
Given #426 and the other open bugs, let's just get this over and done with. It won't affect most linux distributions anyway and the few that may care (RHEL and other LTS) can patch this locally.

whot:
ok, I've had another think about this on the weekend and the answer is - there are plenty of things that would be nice to have for a new API, but given we have very few actual users (libinput and mutter, judging by the fedora spec files) so any actual ABI changes are just busy work, nothing is really needed. We could get rid of the more deprecated APIs like the binary eraser check but really, it doesn't gain us much.

PR #291, opened 2020-10-06
Create xp-pen-g430.tablet

YaLTeR:
Not sure how to create a `.layout` or if it's needed (the tablet has no buttons). Not sure where `Styli` comes from, I took it from `xp-pen-star03.tablet`.

https://www.xp-pen.com/product/52.html

image

### DIFF
--- a/data/xp-pen-g430.tablet
+++ b/data/xp-pen-g430.tablet
@@ -1,0 +1,18 @@
+# XP-Pen
+# G430
+#
+
+[Device]
+Name=XP-Pen G430
+ModelName=
+DeviceMatch=usb:28bd:0075
+Class=Bamboo
+Width=4
+Height=3
+Styli=0xffffd;
+
+[Features]
+Stylus=true
+Reversible=true
+Touch=false
+Buttons=0


whot (changes requested):
the `0xffffd` stylus (see `data/libwacom.stylus`) is the generic pen with no eraser (i.e. it has no serial number or other type bits). for the XP-Pen tablets, that's the right one.

Our CI is currently out of service, but running the check locally: this doesn't pass the tests (`make check` or `ninja -C builddir test`, depending how you build). Looks like the `Class=Bamboo` line is missing, and ideally also an empty `ModelName=` line.

YaLTeR:
> Our CI is currently out of service, but running the check locally: this doesn't pass the tests (`make check` or `ninja -C builddir test`, depending how you build).

Oh, my bad, forgot to run tests.

> Looks like the `Class=Bamboo` line is missing

There are several classes without any explanation of the difference that I could find. Is `Bamboo` the right one?

> and ideally also an empty `ModelName=` line.

This one was also not very clear, looking at other files I'd do something like `Name=XP-Pen` and `ModelName=G430`, but only `Name` is displayed in GNOME Settings making it nicer to put the model there.

whot:
Yes, bamboo is the right one for this tablet. classes are deprecated anyway but we still require them in the data files.

The point of the `ModelName` is to distinguish the real name from the marketing name (e.g. PTH-660 vs the ambiguous "Wacom Intuos Pro"). In this case the model name is part of the name anyway, so it can be left empty.

YaLTeR:
Done.

whot (approved):

whot:
merged, thanks!

PR #293, opened 2020-10-07
travis CI: update our version to Ubuntu 20.04

whot:


whot:
yay, and our travis CI is back in business (after clicking around in circles for about an hour)

PR #295, opened 2020-10-08
When loading devices for a vid/pid, fall back to bluetooth and i2c

whot:
The term 'usbid' here is misleading in the API and it's not worth adding
separate functions (or a new generic function) for the other busses. Where the
vid/pid matches, return the device for any bus (except the vid/pid-less serial
bus).

### DIFF
--- a/test/test-load.c
+++ b/test/test-load.c
@@ -241,6 +241,22 @@ test_bamboopen(struct fixture *f, gconstpointer user_data)
 	libwacom_destroy(device);
 }
 
+static void
+test_isdv4_4800(struct fixture *f, gconstpointer user_data)
+{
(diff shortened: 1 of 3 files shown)


test/test-load.c
```diff
@@ -273,6 +289,9 @@ int main(int argc, char **argv)
 	g_test_add("/load/056a:WACf004", struct fixture, NULL,
 		   fixture_setup, test_wacf004,
 		   fixture_teardown);
+	g_test_add("/load/056a:48000", struct fixture, NULL,
```
Pinglinux, line 292: I guess "/load/056a:48000" should be "/load/056a:4800". Except that, the logic sounds right to me.

whot: fixed, thanks

PR #301, opened 2020-11-12
Replace Travis CI with GitHub Actions

whot:
Travis CI is about to severely limit the free tiers. Specifically, we would
get 1000 minutes (which at current use would last us a few months) but once
they're up, we have to request more on a case-by-case basis.
This is not workable.

Switch to GitHub actions instead. This patch convers the autogen, meson and
build-from-tarball parts over.

It does **not** provide:
- ppc64le
- ~~FreeBSD~~ *edit: implemented as VM run on macos*
- coverity

In the ppc64le ~~and FreeBSD~~ case I don't know how how to get those working, cc @sanjaymsh, @jbeich 
There doesn't seem to be any available runners for either of those.

Coverity is just a matter of porting from travis to the github action syntax.

~~The travis.yml file is removed despite the above missing items, git will
provide its content for future additions without the need to run two CIs.~~ *edit: left in place now*

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -1,0 +1,21 @@
+name: freebsd
+
+on: [ push, pull_request ]
+
+jobs:
+  ###
+  #
+  # FreeBSD build job
+  #
+  freebsd:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: meson test
+      uses: vmactions/freebsd-vm@v0.0.7
+      with:
+        prepare: |
+          pkg install -y meson pkgconf evdev-proto libgudev libxml++ bash
+        run: |
+          meson setup builddir
+          meson test -C builddir --print-errorlogs
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,156 @@
+on: [ push, pull_request ]
+
+env:
+  CFLAGS: "-Werror -Wall -Wextra -Wno-error=sign-compare -Wno-error=unused-parameter -Wno-error=missing-field-initializers"
+  UBUNTU_PACKAGES: libgudev-1.0-dev libxml++2.6-dev valgrind tree
+
+jobs:
+  ###
+  #
+  # autotools build job
+  #
+  autotools:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        compiler:
+          - gcc
+          - clang
+        make_args:
+          - distcheck
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install libwacom dependencies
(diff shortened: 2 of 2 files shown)


whot:
Workflow run for this PR: https://github.com/whot/libwacom/actions/runs/359102163

jbeich:
> FreeBSD

Can you try [workaround](https://github.com/actions/runner/issues/385#issuecomment-702592563) (see [libva example](https://github.com/intel/libva/pull/460/files#diff-6f2adbfa3a8863a041c5953f37db9c434866c90dd11b5b866c23c1c5a87f3e5c)) If it doesn't help I can write a config for CirrusCI or builds.sr.ht.

whot:
> Can you try workaround (see libva example)? If it doesn't help I can write a config for CirrusCI or builds.sr.ht.

yep, that worked, force pushed and [everything ran successfully](https://github.com/whot/libwacom/actions/runs/361007219). So that's BSD sorted then.


Noteworthy: the Actions don't show up here because [according to this post](https://github.community/t/github-actions-are-not-shown-in-the-draft-pr-from-a-fork/17871/2) "A fork can not introduce workflows for the first time and cause the GitHub Actions app to be installed. " for security reasons. Once merged it should work as normal, but for now I'll have to link to my repo.

jigpu:
Would you consider tightening up the test matrix a bit? I know Gtihub grants unmetered access to Actions for public repos, but we could probably trim a few cases without loosing much of anything. I think the following builds are reasonable candidates for the axe:
* The "make check" builds (should already be covered by "make distcheck" case)
* Argument-free Meson builds (Db_sanitize and Valgrind cases would still catch normal non-memory issues)
* Autogen and configure tarball builds (autotools test already runs the former, "make distcheck" runs the latter)

whot:
>   The "make check" builds (should already be covered by "make distcheck" case)

Dropped. There's a slight caveat but it doesn't apply to us here: autotools has a "skip this test on distcheck" variable which I have used in other repos before, so `make check` and `make distcheck` may produce different results. Not the case for us here though.

>  Argument-free Meson builds (Db_sanitize and Valgrind cases would still catch normal non-memory issues)

True, but I really don't want to drop the 99% case from a test suite, even if it should be covered by other tests. So that one I'll leave in, purely for my own psychological benefit ;)

>  Autogen and configure tarball builds (autotools test already runs the former, "make distcheck" runs the latter)

I dropped the configure case now but I left the `autogen.sh` case in place. This test is good for catching errors with `EXTRA_DIST` where `autogen.sh` runs fine in the git tree but fails in the tarball.

whot:
I've left `travis.yml` in place now, we can remove this in the future when the full switch is ready. Meanwhile, let's merge the GH actions because travis is unreliable already. Commits squashed together into one, there's no point keeping the in-development history.

PR #303, opened 2020-11-20
data: Add ISDv4 48eb (XPS 13 2-in-1 9310)

guilhem:


jigpu:
Looks good, though can you fix the typo in the commit title ("USDv4" should be "ISDv4")

guilhem:
Fixed
thanks @jigpu

PR #309, opened 2020-11-27
Enforce lowercase hex for DeviceMatch

whot:
If we have a mix of both, `git grep` doesn't work to check for existing matches, so let's be consistent.


whot:
Sigh, I don't know what's going on here. The BSD job fails and it's frustrating not being able to reproduce this in a local container. @jbeich, can you take a look please? I left the last commit in place which explains most of it, but the basic problem is:

```
sudo pkg install -y meson py37-pip libevdev
sudo pip install libevdev pytest pyudev
python -c 'import libevdev'  # works for the other pip modules too
```
so everything looks fine here. but in meson.build:
```
python.find_installation(modules: ['libevdev'])
```
this fails. Not just for libevdev but also for pyudev (and probably pytest, didn't test) but unfortunately there are no error messages. This *could* be a meson issue, but the same code works on Fedora (locally) and Ubuntu (in travis) and I don't have an environment to test FreeBSD.

jbeich:
> The BSD job fails and it's frustrating not being able to reproduce this in a local container.

Looks like Meson can't find whatever PIP installs. I can't reproduce inside my FreeBSD 12.1 amd64 jail.

From [error log](https://travis-ci.com/github/linuxwacom/libwacom/jobs/449479670):
> ```
> $ python -c 'import sys; print(sys.path)'
> ['', '/home/travis/.pyenv/versions/3.6.10/lib/python36.zip', '/home/travis/.pyenv/versions/3.6.10/lib/python3.6', '/home/travis/.pyenv/versions/3.6.10/lib/python3.6/lib-dynload', '/home/travis/.pyenv/versions/3.6.10/lib/python3.6/site-packages']
> ```

`python` symlink refers to 3.6 while Meson uses 3.7. Let's look closer:

> ```
> $ sudo pkg install -y meson py37-pip libevdev
> ...
> [1/7] Deinstalling py36-setuptools-41.2.0...
> [1/7] Deleting files for py36-setuptools-41.2.0: 100%
> [2/7] Installing py37-setuptools-44.0.0...
> [3/7] Installing ninja-1.10.1,2...
> [3/7] Installing py36-setuptools-44.0.0...
> [4/7] Installing meson-0.55.3...
> [5/7] Installing py37-pip-20.2.3...
> [6/7] Installing libevdev-1.9.1.20200928...
> ```

Try adding `sudo pkg upgrade` before installing `py37-pip` or replace `py37-pip` with `py36-pip`.

I can't test myself due to "Builds have been temporarily disabled for private and public repositories due to a negative credit balance."

whot:
This is on top of #311 now and adjusted for github actions.

@jbeich: thanks for testing. `py36-pip` wasn't available (I did have that at first but had to change). But the same snippet now works on the github actions, so... *shrug*

jigpu (commented):
Ack, though s/Ubunto/Ubuntu/ on the Github Actions commit

PR #326, opened 2021-01-07
data: link to the wacom-hid-descriptors repo in the example tablet file

whot:
Since we expect users to submit the sysinfo data anyway, might as well link to
it so they run that first.


jigpu (approved):
LGTM :) Speaking of that, I really should get around to documenting the maintainer process for wacom-hid-descriptors...

PR #328, opened 2021-01-07
Support loading tablet/stylus files from /etc/libwacom

### LINKED ISSUE #327: Support loading tablet files from /etc/
We load all our data files from `/usr/share/libwacom`. Users adding new devices need to modify a (potentially read-only) system directory, followed by their files potentially being overwritten by a future update (sometimes causing duplicate tablet definitions if the filename changed).

Convention is that system data goes in `/usr/share/`, host-specific configuration in `/etc/` so we should just do the same.

whot:
To add a new .tablet file, a user would have to add that file to /usr/share.
This location may be read-only on some systems, and any file there will
conflict with files provided by the distribution package. In the best case,
the user's file will be overwritten on update, in the worst case a file name
change leaves both files in place, causing conflicts.

The traditional approach is to have distribution-provided files in /usr and
host-specific packages in /etc, so let's do the same here: add /etc/libwacom
as default lookup path for .tablet and .stylus files.

The CI is a bit quirky again because afaict github workflows don't do templates. So it's some duplication but the basic summary is:
- build and install everthing with prefix `/usr`
- generate a list of supported devices with the `./builddir/list-devices` tool
- move some tablet files over to `/etc/libwacom`
- generate a list of supported devices with the `./builddir/list-devices` tool
- compare the two lists, they must be identical

Fixes #327

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -106,6 +106,49 @@ jobs:
             builddir/meson-logs/testlog*.txt
             builddir/meson-logs/meson-log.txt
 
+  ####
+  # /etc/ loading check
+  etcdir:
+    needs: meson
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        command:
+          # A variety of ways to split the database across the two locations
+          # we load from.
+          - sudo mv /usr/share/libwacom/intuos*.tablet /etc/libwacom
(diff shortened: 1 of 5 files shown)


whot:
Merged, it's something we should definitely have to keep in line with other standards.

PR #329, opened 2021-01-08
gaomon-pd1560 tablet configs

AlexWithBeard:


data/gaomon-pd1560.tablet
```diff
@@ -0,0 +1,24 @@
+# GAOMON
+# PD1560
+#
+# This file is a slighlty modified gaomon-s620.tablet
```
whot, line 4: typo: "slightly"

data/gaomon-pd1560.tablet
```diff
@@ -0,0 +1,24 @@
...
+
+[Device]
+Name=GAOMON PD1560
+ModelName=
+DeviceMatch=usb:256c:006e
```
whot, line 9: you'll need the name here as well to match correctly, unfortunately that USB ID is re-used my multiple devices, without the name match we cannot be correct here. See the S620 file for details.

whot (changes requested):
Thanks for the PR, a few minor changes for the matching but this should be fine otherwise.

Please run the `sysinfo.sh` script from [this repo](https://github.com/linuxwacom/wacom-hid-descriptors) and link to the PR in the `.tablet` file (see other files for reference). This will make debugging any issues with the tablet in the future a lot easier.

whot:
Closing, almost 18months with no reply

PR #339, opened 2021-01-29
workflows: run apt update + install instead of apt-get

whot:
Otherwise we are trying to pull down old packages and failing with 404s


whot:
Merging this because the failing CI blocks any other work right now

PR #340, opened 2021-01-29
tools: add a libwacom-show-stylus tool to easily check for stylus IDs

whot:
This is an interactive tool to allow users to easily find their stylus IDs.
libinput debug-tablet provides something similar but it's useful to have this
as part of our own repository.
Example output from running without arguments:

    $ sudo libwacom-show-stylus
    Using "Wacom Intuos Pro M Pen": /dev/input/event23
    Tool id 0x842 serial 0x748012bf in-proximity: False
    Tool id 0x84a serial 0x748012bf in-proximity: False

edit: with the second commit, it now also adds this:
```
Styli=mobilestudio
```
i.e. it loads `libwacom.stylus` and matches the ids with the groups.

cc @jigpu for input please. autotools integration is missing but otherwise this seems like a good addition?

### DIFF
--- a/tools/show-stylus.py
+++ b/tools/show-stylus.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python3
+#
+# Permission to use, copy, modify, distribute, and sell this software
+# and its documentation for any purpose is hereby granted without
+# fee, provided that the above copyright notice appear in all copies
+# and that both that copyright notice and this permission notice
+# appear in supporting documentation, and that the name of Red Hat
+# not be used in advertising or publicity pertaining to distribution
+# of the software without specific, written prior permission.  Red
+# Hat makes no representations about the suitability of this software
+# for any purpose.  It is provided "as is" without express or implied
+# warranty.
+#
+# THE AUTHORS DISCLAIM ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+# INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS, IN
+# NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY SPECIAL, INDIRECT OR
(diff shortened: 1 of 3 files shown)


tools/show-stylus.py
```diff
@@ -0,0 +1,124 @@
...
+    print(
+        "One or more python modules are missing. Please install those "
+        "modules and re-run this tool."
+    )
+    sys.exit(1)
+
+ANSI_UP = "\x1B[%dA"
+ANSI_DOWN = "\x1B[%dB"
+ANSI_RIGHT = "\x1B[%dC"
+ANSI_LEFT = "\x1B[%dD"
```
jigpu, line 42: Would like to see these used instead of the impenetrable raw strings down below

tools/show-stylus.py
```diff
@@ -89,36 +82,75 @@ def main():
...
+        return [s[0] for s in styli.keys()]
+
+
+def load_data_files():
+    stylusfile = Path('./data/libwacom.stylus')
+    if not stylusfile.exists():
+        stylusfile = Path('/usr/share/libwacom/libwacom.stylus')
+    if not stylusfile.exists():
+        die("Unable to find libwacom.stylus")
+    print(f'Using stylus file {stylusfile}')
```
jigpu, line 122: Should probably also add `/etc/libwacom` to the search path now that the library looks there.

jigpu (commented):
Generally seems fine. I doubt the "Suggested" line is going to be perfect, but it should be good enough for patch review.

tools/show-stylus.py
whot: done. I've added the correct behaviour for the `prefix` now too, though there's a side-effect: if the prefix doesn't have files, we load from `/usr/share` anyway. Too niche to worry about, I reckon.

tools/show-stylus.py
whot: done

PR #366, opened 2021-03-15
README: fix the udev instructions

### LINKED ISSUE #364: [help needed] updating instalation to add support for a new tablet to libwacom
Sorry if the question is dumb, but I just spent a good amount of time trying to figure out what to do and could not.

I am using Ubuntu 20.04 with the libwacom version that is shipped by the distro.

Support to my tablet was added a couple of days ago #360 

I have downloaded the .tablet file from /data/ and copied it to /usr/share/libwacom/

According to the readme.md instructions, my next step should be:
**When updating an installed version of libwacom,** add it manually to the existing ruleset (/lib/udev/rules.d/65-libwacom.rules)

I have no idea how to add to the rule set. My device still does not appear when I run libwacom-list-local-devices

whot:
We no longer have a ruleset, we're using the hwdb now - update accordingly.

Fixes #364 (hopefully)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,34 @@ Use the libwacom-list-local-devices tool to list all local devices recognized by
 2. A new tablet description is enabled by adding and installing a new file with a .tablet suffix. Once installed the tablet is part of libwacom's database, no rebuild is neccessary
 3. The tablet is then available through libwacom-list-local-devices
 
-**The new device should also be added to the udev rule to ensure all required properties are set**
-* ***When building from source*** generate an update ruleset with tools/generate-udev-rules after adding the tablet descripton to
-* ***When updating an installed version of libwacom***, add it manually to the existing ruleset (/lib/udev/rules.d/65-libwacom.rules)
+## Updating udev
+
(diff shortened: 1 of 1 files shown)


raphaelchinchilla:
Two suggestions to complement the Readme:
I believe that in Ubuntu the folder is `/lib/udev/hwdb.d/` (without the /etc)
I believe that running xinput returns VendorID:ProductID

whot:
>  I believe that in Ubuntu the folder is /lib/udev/hwdb.d/ (without the /etc)

`/lib/udev` is for the system-provided files, i.e.. in your case the ones from Ubuntu packages. As a general rule, they shouldn't be touched since they will get overwritten on update. `/etc/udev/` is for host-specific configuration, so any custom configuration should go in there.

>  I believe that running xinput returns VendorID:ProductID

It does, but only if you're running Xorg. If you're running Wayland you just get XWayland's emulated devices and they don't have a vid/pid, so it's easier to just point at the file that definitely includes the info.

Thanks though :)

PR #371, opened 2021-03-17
Replace the hwdb generation tool with a python tool

### LINKED ISSUE #365: cross-compilation support
Hey, I was looking at getting cross-compilation working (for aarch64 in this case) for nixpkgs since libwacom is a dependency of one of the packages I'm trying to compile (libinput, which is in turn a dependency of phoc). However, it seems like libwacom doesn't support cross-compilation completely due to a small oversight in the meson.build file; generators (which should run on the build platform) are not tagged with `native: true`, which signals to meson that the build machine compiler and environment should be consulted instead of that of the host (/target) machine.

I'm trying to do this within the nixpkgs repository, but that shouldn't really matter to you.

I've made a patch, but I think it terrible, so I'd like to hear how this could be done in a cleaner way (or if not, I'd like to hear that too). However, it seems like it at least successfully compiles with this patch and the correct dependencies setup, so there's that.

Also: if we can figure out a nice way to do this, are patches for cross-compilation accepted?

Main thing here is that the 'build' and 'host' platforms should be properly distinguished, all the way up the dependency chain (afaict).

<details><summary>The patch in question</summary>

```diff

diff --git a/meson.build b/meson.build
index 32ef797..282d9ab 100644
--- a/meson.build
+++ b/meson.build
@@ -45,9 +45,11 @@ cflags = cc.get_supported_arguments(
 add_project_arguments(cflags, language: 'c')
 
 # Dependencies
-pkgconfig   = import('pkgconfig')
-dep_gudev   = dependency('gudev-1.0')
-dep_glib    = dependency('glib-2.0')
+pkgconfig      = import('pkgconfig')
+dep_gudev      = dependency('gudev-1.0')
+dep_glib       = dependency('glib-2.0')
+build_dep_gudev = dependency('gudev-1.0', native: true)
+build_dep_glib  = dependency('glib-2.0', native: true)
 
 includes_include = include_directories('include')
 includes_src = include_directories('libwacom')
@@ -68,6 +70,10 @@ deps_libwacom = [
            dep_glib
 ]
 
+build_deps_libwacom = [
+           build_dep_gudev,
+           build_dep_glib
+]
 inc_libwacom = [
             includes_include,
             includes_src,
@@ -87,7 +93,22 @@ lib_libwacom = shared_library('wacom',
                                '-DDATADIR="@0@"'.format(dir_data),
                              ],
                              install: true)
+build_lib_libwacom = shared_library('wacom-for-generator',
+                             src_libwacom,
+                             include_directories: inc_libwacom,
+                             dependencies: build_deps_libwacom,
+                             version: libwacom_so_version,
+                             link_args: version_flag,
+                             link_depends: mapfile,
+                             c_args: [
+                               '-DG_LOG_DOMAIN="@0@"'.format(meson.project_name()),
+                               '-DDATADIR="@0@"'.format(dir_data),
+                             ],
+                             install: false,
+                             native: true)
+
 dep_libwacom = declare_dependency(link_with: lib_libwacom)
+build_dep_libwacom = declare_dependency(link_with: build_lib_libwacom)
 
 install_headers('libwacom/libwacom.h', subdir: 'libwacom-1.0/libwacom')
 
@@ -135,10 +156,11 @@ tools_cflags = ['-DTOPSRCDIR="@0@"'.format(meson.source_root())]
 
 gen_hwdb = executable('generate-hwdb',
                      'tools/generate-hwdb.c',
-                     dependencies: [dep_libwacom, dep_glib],
+                     dependencies: [build_dep_libwacom, build_dep_glib],
                      include_directories: [includes_src],
                      c_args: tools_cflags,
-                     install: false)
+                     install: false,
+                     native: true)
 custom_target('hwdb',
              command: gen_hwdb,
              capture: true,

```

whot:
Side-effects:
- cross-compilation is trivial (#365)
- users can run the tool directly after installing a `.tablet` file instead of having to manually write a hwdb file.

Fixes #365


whot:
@jbeich - what do I need to do to make `#!/usr/bin/env python3` work on freebsd? My guess to add `/usr/bin/local` to `$PATH` didn't work :)

jbeich:
@whot, try `pkg install -y python3`

jbeich:
See #373

PR #373, opened 2021-03-18
CI: unhardcode pip and install python3 symlink

jbeich:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=253815 will replace `py37-pip` with `py38-pip`. While `devel/py-pip` will work other Python packages may not be under `devel/` directory (or physical category), so let's use `-g` to enable glob expansion of `*`, `?`, `[`.

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -16,7 +16,7 @@ jobs:
       with:
         prepare: |
           pkg install -y meson pkgconf evdev-proto libgudev libxml++ bash libevdev
-          pkg install -y -g py\*-pip python3
+          pkg install -y -g py3\*-pip python3
           pip install libevdev pytest pyudev
         run: |
           meson setup builddir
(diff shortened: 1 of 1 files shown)


.github/workflows/freebsd.yml
```diff
@@ -15,7 +15,8 @@ jobs:
       uses: vmactions/freebsd-vm@v0.1.2
       with:
         prepare: |
-          pkg install -y meson pkgconf evdev-proto libgudev libxml++ bash libevdev py37-pip
+          pkg install -y meson pkgconf evdev-proto libgudev libxml++ bash libevdev
+          pkg install -y -g py\*-pip python3
```
jbeich, line 19: `py*-pip` won't install `py27-pip` as all `py27-*` are phased out: very few remain, required by other packages. Besides, `py3?-pip` won't match `py310-pip` once Python 3.10 becomes default.

whot, line 19:
```suggestion
          pkg install -y -g py3\*-pip python3
```
We don't need pip 2.7 around, do we?

whot (approved):
Thanks for the quick reply. The pip2.7 is the only question I have here, otherwise this LGTM

whot:
Merged it with the python3 pip change, let's hope it doesn't break anything. If it does, I'll restore the original. :)

Thanks!

PR #376, opened 2021-03-19
data: add a test for receiver IDs

whot:
We don't want any .tablet entries with the receiver ID as that may match
multiple devices. We already test this as part of the tablet-validity test
(see 134166f) but it's a bit more obivous this way.

cc @jigpu


whot:
Note: this was sitting on a 2y old branch here and I'm cleaning out those ;)

PR #381, opened 2021-03-22
Explicitly run the hwdb generator through python

whot:
Punt finding the python installation to the build system

Fixes https://github.com/linuxwacom/libwacom/issues/365#issuecomment-802700694


whot:
cc @Mindavi

Mindavi:
Looks good and prevents having to patch the shebang line. Can't say much about the autotools stuff, but the meson approach looks fine to me. It all seems to build properly now without having to do anything specific for nixpkgs.

Great stuff!

whot:
thanks for testing!

PR #382, opened 2021-03-22
Allow duplicate devices as long as they're in separate data dirs

### LINKED ISSUE #379: not all Huion H950Ps use the same Device ID
While this isnt something you guys can remedy I did feel it was necessary to report it to make it known to you and to the community is that not all H950P devices use the same ID.

Many of the Huion H950P ID with  256c:006d but some like in the case of mine use 256c:006e. This is some weirdness on Huions part but requires select users to have to edit the huion-h950p.tablet and huion-new-1060-plus.tablet to reflect the specific ID for their system so the H950p in question doesnt try to be a 1060+ and you can have the proper button layout.

While it will still work OOTB when it is identified as a 1060+ but setting up buttons can be a bit different . Also of note is that the top and bottom buttons are flipped on the pen in settings for the 256c:006e model of H950P (not sure if this is ALL of them though)

EDIT:

Fixing it for these models you need to swap DeviceMatch= in the new-1060 and the H950P .tablet files

in the H950P file change
`DeviceMatch=usb:256c:006d` -> `DeviceMatch=usb:256c:006e:HID 256c:006e Pen;usb:256c:006e:HID 256c:006e Pad`

and vice versa for the 1060+ file

whot:
This enables a user to copy/paste a `.tablet` file to `/etc/libwacom/`, change the `DeviceMatch=` line and have it magically work, even if the original file still exists in `/usr/share/libwacom`.

Previously we failed with an error about this being a duplicate - that error is still there but only if duplicate matches are used within the same data directory.

Fixes #379

cc @Big-FG

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -153,6 +153,52 @@ jobs:
       - name: compare device database
         run: diff -u8 devicelist.default.txt devicelist.modified.txt
 
+  ####
+  # duplicate device check
+  duplicate-devices:
+    needs: meson
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      # install python so we get pip for meson
+      - uses: actions/setup-python@v1
+        with:
+          python-version: '3.8'
+      # Run as sudo because we install to /etc and thus need the pip
+      # packages available to root
+      - uses: ./.github/actions/pkginstall
+        with:
+          apt: $UBUNTU_PACKAGES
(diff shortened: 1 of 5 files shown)


Big-FG:
Awesome, Ill give this a build and try later today. Appreciate this for working around HUIONs issues

Big-FG:
Just wanted to verify this is working perfectly. I was able to use /etc/libwacom for my tablet file after building master with this MR

whot:
thanks for testing!

PR #389, opened 2021-03-25
Add libwacom_database_ref/unref

whot:
This enables a client to re-use a database context without having to jump
through hoops. Specifically libinput right now wraps the libwacom database
into refcounter struct for that reason - it's easier if we can just use a
libwacom-provided function for this use-case.

This deprecates libwacom_database_destroy(), still works the same as before
assuming the refcount is correct.


whot:
Going to close this one for now. Mutter doesn't need it and libinput cannot use it in this format - it needs to know whether there's any other references left. So this would be churn for little benefit.

PR #393, opened 2021-04-04
CI: unbreak FreeBSD build

jbeich:
Clock desync due to running under a VM confuses Meson. Noticed via https://github.com/intel/libva/commit/fae0a978c304961b670027f644fb9305c009a9bd


whot:
thanks!

PR #394, opened 2021-04-22
tools: use a YAML-compatible format by default

whot:
Let's output something that is both human- and machine-readable. This makes
it easier to search for specific devices with YAML parsers, e.g. to select all
bluetooth devices one could use:
``
  $ libwacom-list-devices | yq '.devices[] | select(.bus == "bluetooth") | .name'
``
Output for list-devices is now:
```
    devices:
    - {bus: i2c, vid: '056a', pid: '5169', name: Wacom ISDv4 5169}
      ...
```
Output for list-local-devices is now for example:
```
    devices:
    - name: 'Wacom Intuos Pro M'
      bus: usb
      vid: '056a'
      pid: '0357'
      nodes:
      - /dev/input/event24
      - /dev/input/event23
      - /dev/input/event22
```


whot:
A few updates here:
- `list-compatible-styli` now also prints out YAML (was kinda the case before already, this just brings it more in-line)
- all strings are quoted now
- vid/pid is now a quoted `0x`-prefixed string, I tried using just hex numbers but especially `yq` ends up using decimal (because json) so it's easier to leave them as quoted string

PR #403, opened 2021-05-28
Add several new tablet definitions

jigpu:


data/isdv4-51e2.tablet
```diff
@@ -1,8 +1,13 @@
...
+# Sensor Type: AES
+# Features: Touch (Integrated), Tilt
+# HW Resolution: 30931 x 17399 (2540 x 2540 lpi)
+#
+# Autogenerated from sysinfo.giGDPrJpNp
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/150#issue-841283965
 
 [Device]
-Name=Wacom ISDv4 51E2
+Name=ISDv4 51e2
```
whot, line 10: what's our preference for these here? quick git grep shows a majority using uppercase letters, only the recent ones use lowercase (probably a `sysinfo.sh` result). Change the new ones to uppercase or change the old ones to lowercase? No preference, but we should be consistent

PR #413, opened 2021-07-16
Prettify our doxygen output

whot:


libwacom/libwacom.h
```diff
@@ -79,24 +79,54 @@
...
 typedef struct _WacomDeviceDatabase WacomDeviceDatabase;
 
+/**
+ * @ingroup devices
+ */
 #define WACOM_STYLUS_FALLBACK_ID 0xfffff
+/**
+ * @ingroup devices
+ */
 #define WACOM_ERASER_FALLBACK_ID 0xffffe
```
jigpu, line 124: I would have expected these to be `@ingroup styli`

jigpu (changes requested):
LGTM aside from one comment.

libwacom/libwacom.h
whot: indeed, thanks. fixed

PR #419, opened 2021-08-09
data: add support for Waltop Slim Tablet 12.1" and rebrands

LeacHLeacH:
This .tablet-file worked for my "Medion Graphics Pad MD 85637" just fine. Nevertheless I realised that it is a re-branded tablet so I used the "original" name. 
I also found out that many companies used this exact model with seemingly the same USB ID but I wasn't able to confirm that.
Further I wasn't able to add support for some software buttons yet. I'm not sure whether this is even possible or not. The tablet itself and the settings I used are working though.

I found the mentioned information [here (Link to DIGImend)](https://digimend.github.io/tablets/Waltop_Slim_Tablet_12.1_inch/).

### DIFF
--- a/data/waltop-slim-tablet-12-1.tablet
+++ b/data/waltop-slim-tablet-12-1.tablet
@@ -13,7 +13,7 @@ ModelName=
 DeviceMatch=usb:172f:0031
 Class=Bamboo
 Width=10
-Height=6.25
+Height=6
 Styli=0xffffd;
 
 [Features]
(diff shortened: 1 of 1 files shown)


LeacHLeacH:
I used the values given by the manufacturer but the build test failed because I used ```Height=6.25```. I'm not quite sure if this is intended behaviour or not.
Is there a reason why the width and height of the tablet have to be an integer and not float?

whot:
>  Is there a reason why the width and height of the tablet have to be an integer and not float? 

It's a marketing measure only and nothing uses it, so just round it to whatever the nearest inch is. Anything that actually uses the tablet will use the real dimensions it gets from the kernel device.

LeacHLeacH:
All right, thanks. I already changed the value which wouldn't fail in e542ece.

whot:
merged, thanks

PR #428, opened 2021-09-24
Add new Intuos BT (CTL-4100WL/CTL-6100WL) device matches

Joshua-Dickens:
Adds the new Intuos BT model ID's into the relevant .tablet files.


whot:
thanks!

PR #429, opened 2021-09-27
Update isdv4-523a.tablet

ajofuk:
Added 523e PID as found in my Yoga 9I 14ITL5. Different PID could be between the HD and 4K display variants? 523e was found on my 4k model. 

Added a new DeviceMatch to this one, rather thank making a new .tablet file as it seems to be the same tablet/laptop. Hope that was right.


whot:
Easier to put it into a separate tablet file, can you do that please. We don't support multiple DeviceMatch lines per file (matches are separated by semicolon instead, see the other files) but if you want the name to show up we need a new file anyway.

whot:
wrong button, sorry.

ajofuk:
Separate tablet file created for isdv-523e, and modifications from the isdv-523a file removed. I'm afraid I'm only using github.com at the moment so was only able to revert the 523a file by manually editing it agin. Hope this is all ok.

whot:
No worries, I squashed them together. Thanks,merged!

PR #437, opened 2021-11-15
Add tablet description for Dell Latitude 5290

CodedOre:
Just recently received a Dell Latitude 5290 and noticed the internal tablet missed a description file.
This is a copy of the file for it's predecessor, the Latitude 5285, with updated device id, making it recognized by, for example, Gnome's settings.


whot:
merged, thanks

Issue #438, opened 2021-11-15
libwacom_new_from_path() fails in mocked environment


hadess:
umockdev doesn't support mocking device nodes, so `libwacom_new_from_path()` fails to find devices after calling `g_udev_client_query_by_device_file()` which always fails.

One possible fix is it use something like `g_udev_client_query_by_subsystem_and_device_file()` (https://gitlab.gnome.org/GNOME/libgudev/-/merge_requests/22) or copy/paste it, so that there's no I/O made at all, and all the data comes from the udev database.

hadess:
Ugly diff. Let me know if that would be a good way to fix this problem, and I'll prepare a PR.

```patch
diff --git a/libwacom/libwacom.c b/libwacom/libwacom.c
index 935344c..f496343 100644
--- a/libwacom/libwacom.c
+++ b/libwacom/libwacom.c
@@ -195,6 +195,22 @@ get_bus (GUdevDevice *device)
        return bus_str;
 }
 
+static GUdevDevice *
+my_udev_client_query_by_device_file (GUdevClient *client,
+                                    const char  *subsystem,
+                                    const char  *path)
+{
+       GList *l;
+       g_autolist(GUdevDevice) devices = NULL;
+
+       devices = g_udev_client_query_by_subsystem (client, subsystem);
+       for (l = devices; l != NULL; l = l->next) {
+               if (g_strcmp0 (g_udev_device_get_device_file (l->data), path) == 0)
+                       return g_object_ref (l->data);
+       }
+       return NULL;
+}
+
 static gboolean
 get_device_info (const char            *path,
                 int                   *vendor_id,
@@ -217,7 +233,7 @@ get_device_info (const char            *path,
        *name = NULL;
        bus_str = NULL;
        client = g_udev_client_new (subsystems);
-       device = g_udev_client_query_by_device_file (client, path);
+       device = my_udev_client_query_by_device_file (client, subsystems[0], path);
        if (device == NULL) {
                libwacom_error_set(error, WERROR_INVALID_PATH, "Could not find device '%s' in udev", path);
                goto out;
@@ -587,7 +603,7 @@ libwacom_new_from_path(const WacomDeviceDatabase *db, const char *path, WacomFal
 out:
        g_free (name);
        if (ret == NULL)
-               libwacom_error_set(error, WERROR_UNKNOWN_MODEL, NULL);
+               libwacom_error_set(error, WERROR_UNKNOWN_MODEL, "unknown model");
        return ret;
[...]

PR #439, opened 2021-11-16
Fix libwacom_new_from_path() failing in mocked environment

hadess:
umockdev doesn't support mocking device nodes, so libwacom_new_from_path()
fails to find devices after calling g_udev_client_query_by_device_file()
that always fails as it cannot stat() the device node to get its type
and major and minor numbers.

Using this reimplementation of g_udev_client_query_by_subsystem_and_device_file()
(see https://gitlab.gnome.org/GNOME/libgudev/-/merge_requests/22),
no I/O is made at all, and all the data comes from the mocked udev database.

Closes: #438

Signed-off-by: Bastien Nocera <hadess@hadess.net>

### DIFF
--- a/libwacom/libwacom.c
+++ b/libwacom/libwacom.c
@@ -195,6 +195,24 @@ get_bus (GUdevDevice *device)
 	return bus_str;
 }
 
+static GUdevDevice *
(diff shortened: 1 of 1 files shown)


whot (approved):

whot:
merged, thanks

PR #445, opened 2021-12-02
Remove autotools, it's meson only now

whot:
This allows us to drop the .gitignore file as well since meson doesn't
dirty the git tree.


jigpu:
Well dangit, guess I can't put off building up muscle memory for meson / ninja any longer :laughing:

whot:
you'll never want to go back to autotools once you get used to meson. just the simple ability to have a builddir by default is amazing :)

eli-schwartz:
Looks like you forgot to delete `libwacom.pc.in`.

PR #451, opened 2021-12-31
Update huion-h610-pro.tablet

Tavlor:
my tablet, a v1 from 2014, presents a different name (HUION H610 Pen & HUION H610 Pad) than the current `DeviceMatch` list.

There is also an additional device that appears in the output of `libwacom-list-local-devices` (HUION H610 Mouse) but I don't think it's needed.


whot:
merged, thanks

Issue #469, opened 2022-01-26
Huion Mini Keydial KD100 


albfan:
device:
![image](https://user-images.githubusercontent.com/220968/151121623-9c2d7df6-0893-492a-a2a3-e56a8231bc4b.png)

receiver:
![image](https://user-images.githubusercontent.com/220968/151121829-d37eadae-0021-4cd3-93b4-538cca1adab5.png)

**Device name**
Huion Mini Keydial KD100 

**Device model identifier**
This is problematic, as this device uses an usb to connect wirelessly the dongle expose:

```
Bus 001 Device 005: ID 256c:006d
```

if device is connected wired it shows:

```
Bus 001 Device 011: ID 256c:006d HUION Huion Tablet_KD100
```

projects like https://github.com/kurikaesu/userspace-tablet-driver-daemon read firmware to detect which device is connected:

```
std::wstring huion_tablet::getDeviceFirmwareName(libusb_device_handle *handle) {
    auto *buffer = new unsigned char[200];
    memset(buffer, 0, 200);

    // Extract the firmware name
    int descriptorLength = libusb_get_string_descriptor(handle, 0xc9, 0x0409, buffer, 130);
```

```
std::string huion_tablet::getDeviceNameFromFirmware(std::wstring firmwareName) {
    // HUION tablets
    if (firmwareName == L"HUION_T188_180718") {
        return "Huion WH1409 v2";
    } else if (firmwareName == L"HUION_T191_190619") {
        return "Huion H1161";
    } else if (firmwareName == L"HUION_T153_160524") {
        return "Huion WH1409 (2048)";
    } else if (firmwareName == L"HUION_T200_210309") {
        return "Huion KD100 mini Keydial";
```

Is it possible to add this matching style?

**udevadm info output**

It exposes js0, mouse* input* but about event:

```
event17 -> ../../devices/pci0000:00/0000:00:14.0/usb1/1-1/1-1:1.0/0003:256C:006D.0024/input/input98/event17
event18 -> ../../devices/pci0000:00/0000:00:14.0/usb1/1-1/1-1:1.0/0003:256C:006D.0024/input/input99/event18
event19 -> ../../devices/pci0000:00/0000:00:14.0/usb1/1-1/1-1:1.1/0003:256C:006D.0025/input/input101/event19
event20 -> ../../devices/pci0000:00/0000:00:14.0/usb1/1-1/1-1:1.1/0003:256C:006D.0025/input/input102/event20
[...]

PR #470, opened 2022-01-28
Add support for Huion MiniDial KD100

albfan:
- [ ] Ring is still not correctly configured ()
- [x] mouse button (was just another button)
- [x] This only covers KD100 connected by wire. 
- [ ] Wireless do not match usb filter (it only shows 256c:006d)
- [x] Only eight buttons are working (using GNOME). Fix rest of buttons in digimend driver

relates #469

### DIFF
--- a/data/huion-kd100.tablet
+++ b/data/huion-kd100.tablet
@@ -1,0 +1,56 @@
+# Huion
+# KD100
+#
+# Button Map:
+# (A=1, B=2, C=3, ...)
+# 
+#   S
+#
+# *---------*
+# |         |
+# | A B C D |
+# |         |
+# | E F G H |
+# |         |
+# | I J K L |
+# |         |
+# | M N O P |
+# |       P |
+# | QQQ R P |
+# |         |
+# *---------*
+#
+#
+# Ring:
+# (a=strip min, A=strip max)
+#
+#   L R
+#
+# *---------*
+# |         |
+# |         |
+# |         |
+# *---------*
+#
+
+[Device]
+Name=Huion KD100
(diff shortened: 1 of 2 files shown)


albfan:
Found how to detect third layer of buttons (9-12) https://gitlab.freedesktop.org/libinput/libinput/-/issues/730

albfan:
Fourth and Fifth layer probably needs something on digimend driver

whot (approved):
Looking through this: the data file itself and SVG appear fine. Since libwacom doesn't control the actual functionality of the tablet, I think we're good to merge this?

albfan:
buttons are correctly reported, fixing degimend driver

albfan:
Mouse button is reported too as a pad button "ButtonS"

So all pad buttons are recognized not and shown in svg

data/huion-kd100.tablet
```diff
@@ -0,0 +1,56 @@
...
+# *---------*
+# |         |
+# |         |
+# |         |
+# *---------*
+#
+
+[Device]
+Name=Huion KD100
+DeviceMatch=usb:256c:006d:HUION Huion Tablet_KD100 Keyboard;usb:256c:006d:HUION Huion Tablet_KD100 Pad;usb:256c:006d:HUION Huion Tablet_KD100 Consumer;usb:256c:006d:HUION Huion Tablet_KD100 Pen;usb:256c:006d:HUION Huion Tablet_KD100
```
whot, line 38: this isn't the right match, you don't want libwacom to match on all event nodes this device produces, only on the one that sends the actual events.

albfan: Got it now

whot:
AIUI this is good enough for the libwacom bits, the kernel drivers are separate so let's merge this!

PR #475, opened 2022-02-11
test: add a test for matching the button count with the evdev codes

whot:
Slightly superfluous since all the other tests will fail anyway thanks
to the libwacom database check but this is a more confined test and
possibly easier to understand in the output.

cc @albfan

### DIFF
--- a/test/test_data_files.py
+++ b/test/test_data_files.py
@@ -69,3 +69,18 @@ def test_svg_exists(tabletfile):
 
     except KeyError:
         pass
+
+
+def test_button_evcodes(tabletfile):
+    config = configparser.ConfigParser(strict=True)
+    # Don't convert to lowercase
+    config.optionxform = lambda option: option
+    config.read(tabletfile)
+
+    try:
+        nbuttons = int(config['Features']['Buttons'])
+        str = config['Buttons']['EvdevCodes']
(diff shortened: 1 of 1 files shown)


albfan:
I see do a test in meson for config files, but the use case would be for users changing or adding config at runtime.

app consuming libwacom should get a null or similar at runtime not something that breaks, and yes, that app should not break, but if output from libwacom is reliable problably can trust that and not protect from errors

whot:
if the parsing fails, libwacom prints an error (`set_button_codes_from_string`) and falls back to heuristics. Which will then print out another warning if they can't match the button (`set_button_codes_from_heuristics`).  In that case `libwacom_get_button_evdev_code()` returns zero (as documented) for that button and it's up to the caller to handle that.

IOW, libwacom already does the documented behavior, this test merely ensures that we have a nicer error message in the CI that's easier to chase down. Nothing else is needed in libwacom, the crash you noticed is a bug in the caller.

albfan:
So for runtime we need a protection on libinput or gnome session? I expect it to be libinput

whot:
> So for runtime we need a protection on libinput or gnome session? I expect it to be libinput

quick skim shows the libinput code to do the right thing - if that's not the case please file a bug against libinput.

PR #480, opened 2022-02-21
Add XP Pen Deco mini7

versusvoid:
Link to the tablet: https://www.xp-pen.com/product/613.html

This at least allows to enable left-handed mode.
It also has 8 pad buttons, but they are handled by the kernel as separate keyboard device.

### DIFF
--- a/data/xp-pen-deco-mini7.tablet
+++ b/data/xp-pen-deco-mini7.tablet
@@ -11,7 +11,7 @@ DeviceMatch=usb:28bd:0928:UGTABLET 6 inch PenTablet
 Class=Bamboo
 Width=7
 Height=4
-Styli=0xffffd;
+Styli=@generic-no-eraser;
 
 [Features]
 Stylus=true
--- a/data/xp-pen-deco-mini7.tablet
+++ b/data/xp-pen-deco-mini7.tablet
@@ -1,0 +1,23 @@
+# XP-Pen
+# Deco mini7
+#
+# sysinfo.KmirkLjsqr.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/268
+
+[Device]
+Name=XP-Pen Deco mini7
+ModelName=
+DeviceMatch=usb:28bd:0928:UGTABLET 6 inch PenTablet
+Class=Bamboo
+Width=7
+Height=4
+Styli=0xffffd;
+
+[Features]
+Stylus=true
+Reversible=true
+Touch=false
+Buttons=8
+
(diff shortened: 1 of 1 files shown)


data/xp-pen-deco-mini7.tablet
```diff
@@ -0,0 +1,18 @@
...
+# Deco mini7
+#
+
+[Device]
+Name=XP-Pen Deco mini7
+ModelName=
+DeviceMatch=usb:28bd:0928:UGTABLET 6 inch PenTablet
+Class=Bamboo
+Width=7
+Height=4.37
```
whot, line 11: this needs to be an integer

data/xp-pen-deco-mini7.tablet
```diff
@@ -0,0 +1,18 @@
+# XP-Pen
+# Deco mini7
+#
```
whot, line 3: missing the sysinfo, see e.g. 9d61a6c0cf4c0052a53afa288a4bf900e4eb2018 on how to add this

whot (commented):
> It also has 8 pad buttons, but they are handled by the kernel as separate keyboard device.

That's actually fairly standard, the stylus and the "Pad" itself have different event nodes. If that's all there is to it, then the buttons should be added to the data file too please.

data/xp-pen-deco-mini7.tablet
```diff
@@ -0,0 +1,23 @@
...
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/268
+
+[Device]
+Name=XP-Pen Deco mini7
+ModelName=
+DeviceMatch=usb:28bd:0928:UGTABLET 6 inch PenTablet
+Class=Bamboo
+Width=7
+Height=4
+Styli=0xffffd;
```
whot, line 14:
```suggestion
Styli=@generic-no-eraser;
```

whot:
Apologies, I missed the updates on this one. Merging now

whot:
Ok, I was too early, this fails the tests with:
```
 Unable to determine evdev code for button E (XP-Pen Deco mini7)
```
What button codes does this device produce? `libinput record` (or `evemu`/`evtest` will tell you). Anyway, I'm assuming it's the same set of codes the MW uses, so I copied the line from there.

The layout file is missing too, so I hacked around this now in #576 by re-using the one from the Deco MW. Feel free to add an svg file for this tablet, it should be doable by copying the svg for the MW and moving the buttons around a bit.

Meanwhile, closing this MR, it's been replaced by #576

PR #481, opened 2022-03-06
Added Fujitsu T935 so gnome-control-center can calibrate the stylus.

frohro:
Hi,
Here is a pull request for the changes that seemed to fix my Fujitsu Lifebook T935, so libwacom-list-local-devices found it, and so the gnome-control-center would allow me to calibrate it.  
I'm a newbie at this, so I hope everything is okay.  I changed the Fujitsu T902 to have a button, because the T935 stylus has one.  I tested that I could control that button in gnome-control-center, and it worked.  The calibration worked as well.  And I don't do pull requests often either, so hopefully that is okay as well.
Thanks,
Rob


whot (commented):
Looks good, thanks. Only thing missing: the sysinfo link which you'll get by running the scripts from https://github.com/linuxwacom/wacom-hid-descriptors/ (including filing the MR there). Let's attach this please, makes it easier to test these devices in the future.

frohro:
There is no set of descriptors for the Fujitsu T935.  What am I supposed to do in this case?  
Thanks,
Rob

whot:
> There is no set of descriptors for the Fujitsu T935. What am I supposed to do in this case?

the sysinfo.sh script didn't produce anything?

frohro:
Okay, I understand better now.  I ran the sysinfo.sh and it produced a file (attached here) and I submitted an issue to their repository with the same file.   [sysinfo.hEgTzdfzP1.tar.gz](https://github.com/linuxwacom/libwacom/files/8215832/sysinfo.hEgTzdfzP1.tar.gz) 
Is that all you needed?
Thanks,

Rob

data/isdv4-10g.tablet
```diff
@@ -0,0 +1,15 @@
...
+DeviceMatch=usb:056a:5019
+Class=ISDV4
+Width=11
+Height=6
+IntegratedIn=Display;System
+
+[Features]
+Stylus=true
+Touch=true
+Buttons=1
```
whot, line 15: `Buttons` refers to the number of buttons on the **tablet**, not on the stylus - is this correct? Quick google image search for the T935 doesn't show any buttons.

data/isdv4-10g.tablet
```diff
@@ -0,0 +1,15 @@
+# this is for the Wacom pen + touchscreen as found in the Fujitsu T935
```
whot, line 1: Have a look at e.g. commit ddc4e01ad61681764af06ef665f6c4dfaba7b3fe, it shows how the sysinfo is mentioned in the .tablet file so we have the link back to it when we need. Once that's in, the test will pass.

frohro: I added the comment about the sysinfo file, committed and pushed it.  I hope this is what you meant me to do.  :-)  Thanks for all the help with this.
Rob

jigpu:
Accidentally filed #484 with the autogenerated tablet data... I've closed that request to let @frohro's version take priority :)

whot:
Fixed with #484 now

PR #489, opened 2022-04-11
CI: pin the meson version to 0.61.4

whot:
0.62.0 fails meson dist with a python complaint, which has been fixed
upstream. But until pip pulls down a fixed version, let's pin meson.

See issue https://github.com/mesonbuild/meson/issues/10181


eli-schwartz:
FWIW, in similar cases (a version of pip that installed Windows binaries with a broken sys.stderr that broke Meson in CI) I've used pinning of `pip != broken-version` with the assumption that once the fixed version is released it will update on its own, but otherwise only skip the known-broken version. So `meson!=0.62.0` would install either 0.61.4 or 0.62.1 and the latter would be fixed.

I don't suppose it matters all that much... :)

whot:
TIL about `!=`, never thought about that. Thanks! I already filed #490 so we don't forget, let's hope this is good enough for *someone* to remember... :)

PR #492, opened 2022-04-22
data: Add ISDv4 5093 (Lenovo ThinkPad X1 Yoga 1st)

jigpu:
Link: https://github.com/linuxwacom/wacom-hid-descriptors/issues/220


whot (approved):

jayaddison:
I have an X1 Yoga 1st edition system that seems similar to the data provided here, with a small difference in the Wacom input USB ID - it uses `056a:5094` instead of `056a:5093`.

Copying the definition file here generally works well, apart from two items:

- Only one button is detected on the stylus (in more detail: the stylus appears to be detected as [`0x11`](https://github.com/linuxwacom/libwacom/blob/cb36c462763a321454d5c08fe974a3d7dec4ed1a/data/libwacom.stylus#L32-L39), which has `Buttons=1`; if I update that to `Buttons=2`, I can use both)
- The eraser does not appear to be functional

Should I provide a `sysinfo.sh` output?

whot:
sysinfo would be appreciated, just in case. thanks

jayaddison:
Thanks, I'll provide that info soon.

(I'm puzzling about why the `Buttons=1` from `0x11` takes precedence over the `Buttons=2` from the `.tablet` definition file.. but perhaps more system info will help to figure that one out)

whot:
buttons in the `.tablet` file refer to the buttons on the tablet device itself (which in your case is 0 since it's an integrated tablet, but think of an Intuos-style tablet instead). The count of buttons for a stylus comes from the `libwacom.stylus` description fil - as you already pointed out above. Can we move this to a separate issue please? if there's something that needs fixing it's better to have it in its own issue than hidden in a PR for some other device.

PR #499, opened 2022-06-18
Add support for Huion Inspiroy H1060P

2talltech:
I thought this was a "Huion New 1060 Plus", but it is newer than that even. It looks exactly like it, but I believe the older model is discontinued. This is initially recognized as the h950p, I assume because that tablet description file matches anything "usb:256c:006d". That model has 8 buttons and this has 12. I cannot make the 16 soft keys at the top work with my limited knowledge.

I can confirm that with these files, I can remap all 12 buttons in Gnome's Wacom Tablet settings with Fedora 36.


data/huion-h1060p.tablet
```diff
@@ -0,0 +1,24 @@
+# HUION
+# H1060P
```
whot, line 2: missing the sysinfo link (see e.g. commit 9d61a6c0cf4c0052a53afa288a4bf900e4eb2018) for how to add this, other than that this looks good, thanks!

2talltech:
I think I have it right now. I also left some breadcrumbs in the `Readme.md` for the next guy who doesn't know where to put the layout file.

whot:
> Bail out! ERROR:../test/test-tablet-svg-validity.c:115:check_button: assertion failed: (node != NULL)

There's a bug in the SVG file, looks like something isn't named correctly, can you check this please?

whot:
> There's a bug in the SVG file, looks like something isn't named correctly, can you check this please?

Sorry, my fault, this appears to be #504, fixed now. Can you please rebase and push so the CI runs again. Thanks

whot:
Superseded by #506 which has the same content but a cleaned-up commit set. Thanks!

PR #508, opened 2022-07-26
Add Chuwi MiniBook X

jibsaramnim:
Link: linuxwacom/wacom-hid-descriptors#241

This PR adds the configuration file necessary for the Chuwi MiniBook X' built-in touch panel to show up in the GNOME Settings Wacom panel.

One minor note; I have picked a Stylus ID that functionally matches the actual pen (Chuwi's H7 pen), but I am not sure if there is a more correct way of selecting this. With the more generic choices the configuration screens were showing two buttons, but the H7's secondary pen button only works as an eraser switch. Please let me know if there is another stylus I should pick, or if there is some kind of command I can run to help provide relevant information of this specific pen if that isn't already in the descriptors shared separately.

Thank you!

### DIFF
--- a/data/chuwi-minibookx.tablet
+++ b/data/chuwi-minibookx.tablet
@@ -1,5 +1,6 @@
 # This is for the CHUWI MiniBook X
 #
+# sysinfo.CSujtSF02A
 # https://github.com/linuxwacom/wacom-hid-descriptors/issues/241
 
 [Device]
--- a/data/chuwi-minibookx.tablet
+++ b/data/chuwi-minibookx.tablet
@@ -5,7 +5,7 @@
 
 [Device]
 Name=Chuwi Minibook X
-Class=ISDV4-aes
+Class=ISDV4
 DeviceMatch=i2c:27C6:011A
 Width=23
 Height=14
--- a/data/chuwi-minibookx.tablet
+++ b/data/chuwi-minibookx.tablet
@@ -7,7 +7,7 @@
 [Device]
 Name=Chuwi Minibook X
 Class=ISDV4
-DeviceMatch=i2c:27C6:011A
+DeviceMatch=i2c:27c6:011a
 Width=9
 Height=6
 IntegratedIn=Display;System;
--- a/data/chuwi-minibookx.tablet
+++ b/data/chuwi-minibookx.tablet
@@ -1,0 +1,17 @@
+# This is for the CHUWI MiniBook X
+#
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/241
+
+[Device]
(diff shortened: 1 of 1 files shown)


data/chuwi-minibookx.tablet
```diff
@@ -0,0 +1,17 @@
+# This is for the CHUWI MiniBook X
+#
```
whot, line 2:
```suggestion
#
# sysinfo.CSujtSF02A
```

data/chuwi-minibookx.tablet
```diff
@@ -0,0 +1,17 @@
...
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/241
+
+[Device]
+Name=Chuwi Minibook X
+Class=ISDV4-aes
+DeviceMatch=i2c:27C6:011A
+Width=23
+Height=14
```
whot, line 10: is that correct? The width and height are in inches (unfortunately), so this seems ... too big for something called minibook.

data/chuwi-minibookx.tablet
```diff
@@ -0,0 +1,17 @@
+# This is for the CHUWI MiniBook X
+#
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/241
+
+[Device]
+Name=Chuwi Minibook X
+Class=ISDV4-aes
```
whot, line 7:
```suggestion
Class=ISDV4
```

data/chuwi-minibookx.tablet
```diff
@@ -0,0 +1,17 @@
+# This is for the CHUWI MiniBook X
```
whot, line 1:
```suggestion
# CHUWI
# MiniBook X
```

data/chuwi-minibookx.tablet
jibsaramnim: Oh gosh, my apologies. I thought it was in centimeters. Converted this would be:

```suggestion
Width=9
Height=6
```

(Note: Technically 14cm is about 5.5", but I think you'd probably prefer to round that up)

whot: yeah, unfortunately here we use inches. Rounding up is fine, it's not like anyone actually uses this value :) Thanks for the quick update.

jibsaramnim: My pleasure! Thanks for taking a look at this PR, I really appreciate it!

jibsaramnim:
I noticed the tests were failing due to me not writing the DeviceMatch in all-lowercase. I've pushed a fix for this out just now, hopefully I didn't miss anything else :).

whot:
Merged, thanks!

PR #520, opened 2022-09-01
Add tablet for Dell Inspiron 14 7425 2-in-1

luyatshimbalanga:
This commit adds support for Dell Inspiron 14 7425 2-in-1.

### DIFF
--- a/data/isdv4-49c8.tablet
+++ b/data/isdv4-49c8.tablet
@@ -1,0 +1,19 @@
+# Dell Inspiron 7425 14 2-in-1
+# Sensor Type: AES (
+# Features: Touch (Intergrated), Tilt
+#
+
+[Device]
+Name=Wacom HID 49C8
+ModelName=
+DeviceMatch=i2c:056a:49c8
+Class=ISDV4
+Width=11
+Height=7
+IntegratedIn=Display;System
+Styli=@isdv4-aes;
+
+[Features]
+Stylus=true
+Touch=true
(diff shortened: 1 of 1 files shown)


luyatshimbalanga:
Also see [this ticket]https://github.com/linuxwacom/wacom-hid-descriptors/issues/251#issue-1358146127)

data/isdv4-49c8.tablet
```diff
@@ -0,0 +1,19 @@
+# Dell Inspiron 7425 14 2-in-1
+# Sensor Type: AES (
+# Features: Touch (Intergrated), Tilt
+#
```
whot, line 4: Please add the comments listing + linking to the sysinfo, like e.g. in de8d5883b61346d0a997c347640498e310d30231, thanks

whot (changes requested):

data/isdv4-49c8.tablet
luyatshimbalanga: Before doing the update, will it be correct to list something like _Dell Inspiron 14 7425 2-in-1 (Wacom HID 49C8)_ as name of device?

whot: everything else in the file is good as it is, the `Name` field should match the kernel's name (doubly so since the same device may be used in other devices too).

whot (approved):

whot:
merged, tahnks!

PR #522, opened 2022-09-08
Add support for a new "Intuos Pro Small"

skomra:
Except PID, everything is the same.

Signed-off-by: Ping Cheng <ping.cheng@wacom.com>


whot:
merged, thanks!

PR #525, opened 2022-09-28
data: Added support for "Cintiq Pro 27"

Joshua-Dickens:
To fix issues with pen compatibility on the new tablet with current libwacom.stylus groups the intuos5 group had certain pens split off into a new group called propengen2. All .tablet files affected by this change were updated to include this new group.

Signed-off-by: Joshua Dickens <joshua.dickens@wacom.com>

### DIFF
--- a/data/intuos-pro-2-s.tablet
+++ b/data/intuos-pro-2-s.tablet
@@ -32,13 +32,13 @@
 [Device]
 Name=Wacom Intuos Pro S
 ModelName=PTH-460
-DeviceMatch=usb:056a:0392;bluetooth:056a:0393;usb:056a:03dc;bluetooth:056a:03dd;
+DeviceMatch=usb:056a:0392;bluetooth:056a:0393;
 Class=Intuos5
 Width=6
 Height=4
 Layout=intuos-pro-2-s.svg
 IntegratedIn=
-Styli=@intuos5;@mobilestudio;
+Styli=@intuos5;@mobilestudio;@propengen2;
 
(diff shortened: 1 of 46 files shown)


data/intuos-pro-2-s.tablet
```diff
@@ -32,13 +32,13 @@
 [Device]
 Name=Wacom Intuos Pro S
 ModelName=PTH-460
-DeviceMatch=usb:056a:0392;bluetooth:056a:0393;usb:056a:03dc;bluetooth:056a:03dd;
+DeviceMatch=usb:056a:0392;bluetooth:056a:0393;
```
Pinglinux, line 35: @Joshua-Dickens You probably accidentally removed the newer Intous Pro Small

Joshua-Dickens: Seems like I did, I think I know what happened. Fixed in a quick commit.

data/libwacom.stylus
```diff
@@ -291,6 +291,22 @@ Buttons=2
...
 
+[0x200]
+# Cintiq Pro 2022
+Name=Pro Pen 3
+Group=cintiqpro2022
+Buttons=3
+Axes=Tilt;Pressure;Distance;
+Type=General
+
+[0x04200]
```
Pinglinux, line 302: @Joshua-Dickens We added an extra 0 at the 4th nibble position for the actual ID in the kernel driver. So, the correct ID here should be 0x40200.

PR #526, opened 2022-10-02
Add Kamva Pro 12 

renk-0:
This `.tablet` file works with the Kamva Pro 12, not shure about the Pro 13.
Also the vertical strip isnt working, but `evtest` and `libinput record` report ABS_WHEEL events not buttons.

### DIFF
--- a/data/kamvas-pro-13.tablet
+++ b/data/kamvas-pro-13.tablet
@@ -1,6 +1,6 @@
 # Huion
-# Kamvas Pro 13
-# GT-133
+# Kamvas Pro 12/13
+# GT-116/GT-133
 #
 # sysinfo.6ktY6Uln9z.tar.gz
 # https://github.com/linuxwacom/wacom-hid-descriptors/issues/160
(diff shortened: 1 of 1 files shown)


data/kamvas-pro-13.tablet
```diff
@@ -30,15 +30,26 @@
...
 # there is nothing we can do about this.
 
 [Device]
-Name=Huion Kamvas Pro 13
-ModelName=GT-133
+Name=Huion Kamvas Pro 12/13
+ModelName=GT-116/GT-133
 Class=Cintiq
-DeviceMatch=usb:256c:006e:Tablet Monitor Pen;usb:256c:006e:Tablet Monitor Pad;
+DeviceMatch=usb:256c:006e:Tablet Monitor Pen;usb:256c:006e:Tablet Monitor Pad;usb:256c:006e:Tablet Monitor
```
whot, line 52: To clarify here: does the device *only* need the new `Tablet Monitor` match or does it require the other ones too?

If the new match alone is sufficient, it's better to have a separate data file here please.

whot: got confused myself reading this again, so let's try to clarify: does your device match *only* the newly introduced "tablet monitor" device, or does it also export the various `Pen`, `Pad` devices?

And if not, did you update the kernel recently, because the `Pen` etc suffix is appended by the kernel driver, so missing that means something's amiss there. There was a libinput bug recently with a similar issue but I can't find ti right now, let's put the bat signal out for @JoseExposito

JoseExposito: I think that you refer to this PR:
https://github.com/linuxwacom/libwacom/pull/542#issuecomment-1373428174

The same advice applies in this case. Try to uninstall DIGImend if you re using it and update your kernel. When using a recent kernel, DIGImend actually removes features rather than adding them.

whot: Thanks! I'm going to close this one for now and if the actual issue is different, we can re-open it.

PR #538, opened 2022-12-17
Added support for XP-Pen Deco01 v2

mwaddoups:
I've added support for the XP-Pen Deco01 v2.

The SVG is just a copy of the existing XP-Pen Deco L as the layout is essentially the same.

Note to use this you currently need a special version of the DigiMEND driver to support this - details on this in the libinput issue [here](https://gitlab.freedesktop.org/libinput/libinput/-/issues/839). Those changes should be merged to the kernel in the new year.

The relevant sysinfo issue is https://github.com/linuxwacom/wacom-hid-descriptors/issues/276.

Thanks for all the work you do!


whot:
merged, thanks!

PR #542, opened 2023-01-05
Fix Huion New 1060 Plus

real-or-random:
Apparently the "Pen" string has changed. 

I'm sure this happened quite a while ago but I can't pinpoint it.

Before commit (tablet not shown in GNOME Settings):
```
/dev/input/event29 is a tablet but not supported by libwacom
/dev/input/event28 is a tablet but not supported by libwacom
/dev/input/event26 is a tablet but not supported by libwacom
devices:
- name: 'Huion New 1060 Plus'
  bus: 'usb'
  vid: '0x256c'
  pid: '0x006e'
  nodes: 
  - /dev/input/event27
```

After commit (tablet shown in GNOME Settings):
```
/dev/input/event29 is a tablet but not supported by libwacom
/dev/input/event28 is a tablet but not supported by libwacom
devices:
- name: 'Huion New 1060 Plus'
  bus: 'usb'
  vid: '0x256c'
  pid: '0x006e'
  nodes: 
  - /dev/input/event27
- name: 'Huion New 1060 Plus'
  bus: 'usb'
  vid: '0x256c'
  pid: '0x006e'
  nodes: 
  - /dev/input/event26
```

### DIFF
--- a/data/huion-new-1060-plus.tablet
+++ b/data/huion-new-1060-plus.tablet
@@ -5,7 +5,7 @@
 [Device]
 Name=Huion New 1060 Plus
 ModelName=
-DeviceMatch=usb:256c:006e:HID 256c:006e Pen;usb:256c:006e:HID 256c:006e Pad
+DeviceMatch=usb:256c:006e:HID 256c:006e;usb:256c:006e:HID 256c:006e Pad
 Class=Bamboo
 Width=10
 Height=6


whot:
The `Pen` part of this is usually appended by the kernel driver, so if it's missing now it's likely some kernel-level change. Have you changed to custom drivers recently? Can you try an older kernel to check this please?

cc @bentiss and @JoseExposito  in case that rings a bell

JoseExposito:
@real-or-random, are you using DIGImend?

There were some accidental changes in this prefix:

- Up to version 9 of DIGImend, the "Pen" suffix was used, current status in libwacom.
- DIGImend v10 removed accidentally the device suffix.
- I upstreammed a fix to the mainline kernel: https://github.com/torvalds/linux/commit/0977fda0587cbc5403651ba169e264aa01e8a026

The desired status is having the "Pen" suffix.  All the code present in DIGImend is now available in the upstream kernel so, depending on your kernel version, you won't need to install any external driver and, hopefully, the problem with the suffix should be fixed.

real-or-random:
> If you are using DIGImend, you can try to uninstall it and see if the fix is already available in your distro.

Thanks! For some reason I had the out-of-tree drivers installed. Removing them indeed fixed the problem.

Issue #515, opened 2022-08-03
Pytests fails 2.4.0 if systemd-hwdb is missing


motorto:
Trying to build from source the libwacom-2.4.0, but the 8th test `8/8 libwacom:all / pytest`  keeps failling without giving much help.

x86_64-glibc voidlinux

Below is the testlog.txt (I have removed the output of the other tests and only kept the 8th one).

I dont know what else to provide, so if I am missing something please tell me so.
<details> 
  <summary>testlog.txt</summary>

```text
Log of Meson test suite run on 2022-08-03T12:45:44.604501

[...]

PR #543, opened 2023-01-11
tests: fix failure if system doesn't use systemd

mhmdanas:
Fixes https://github.com/linuxwacom/libwacom/issues/515.

Edit: the CI failure seems to be an unrelated issue.


whot:
> Edit: the CI failure seems to be an unrelated issue.

yep, some server being upset during the run, re-triggered and it passes now, thanks!

PR #549, opened 2023-02-15
Add Wacom HID 52D5 (Lenovo Yoga 7 14ARB7)

stuarthayhurst:
I'm not sure about the height and width, as I don't know the units the file wants, but I'll happily fill this in if you have any pointers. It's 302x188mm according to `libinput`.

Let me know if anything else is missing :)


whot:
>  I don't know the units the file want

inches please, whatever is closest. that is the marketing with (stemming from things like the Intuos sold as 9x6 or whatever). Don't think anyone actually uses this, certainly not for built-in ones.

whot:
Merged, thanks

PR #308, opened 2020-11-26
Add support for Huion H950P

jconnolly837:
This should add support for the Huion H950P tablet and matching pen. 

Fixes #307

PR #562, opened 2023-04-04
data: narrow down the Huion H950p to include a name match

### LINKED ISSUE #556: GAOMON S620 is detected as Huion H950P in GNOME Settings
The issue is caused by the generated `/usr/lib/udev/hwdb.d/65-libwacom.hwdb` file that's missing these lines under `# GAOMON S620`:

```
libwacom:name:* Dial:input:b0003v256Cp006D*
 ID_INPUT_TABLET_PAD=1

libwacom:name:* Touch Strip:input:b0003v256Cp006D*
 ID_INPUT_TABLET_PAD=1
```

See more about this issue: GNOME Settings issue [#2018](https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/2018).

whot:
Huion re-uses the USB IDs for, presumably, fun and profit and the only way we can distinguish them is with name matches. To make things even more fun, *other* vendors like Gaomon re-use Huion's VID/PID too.

For all devices except the H950p we match on name, but this one was the first one introduced so it only matches on VID/PID. As a result, all devices that don't match on the name get matched for the H950p - including the Dial and Touch Strip subdevices of the Gaomon S620 (see #556). For the user this looks like they have a Gaomon device (pen and pad) and a Huion H950p (dial and touchstrip).

Fix this by narrowing down the H950p to the device name too. The name comes from the sysinfo file linked to in the data file.

cc @jconnolly837 from #308

cc @JoseExposito because he *may* have a H950p or touched the kernel patches for that.1

Fixes #556

### DIFF
--- a/data/huion-h950p.tablet
+++ b/data/huion-h950p.tablet
@@ -6,7 +6,7 @@
 [Device]
 Name=Huion H950P
 ModelName=
-DeviceMatch=usb:256c:006d
+DeviceMatch=usb:256c:006d:HID 256c:006d Pen;usb:256c:006d:HID 256c:006d Pad
 Class=Bamboo
 Width=9
 Height=5


JoseExposito:
@whot I have the H640p, just in case you need me to test something with it.

whot:
> I have the H640p, just in case you need me to test something with it.

Ah, cool. That one already has the name match, so the H950p is the only one that matters for this MR. Thanks though!

jconnolly837:
Hey! From what I gather, you already have what you need. I do still have the H950p packed away and a degraded knowledge of git just in case anything absolutely critical comes up.

whot:
@jconnolly837 - luckily in this case there's no git knowledge needed, you can download the file directly [here](https://raw.githubusercontent.com/linuxwacom/libwacom/38f165e77713075f313ab548b203e7a9fafafe85/data/huion-h950p.tablet) (a lot of click throughs to get to the raw file as of this version...) and replace `/usr/share/libwacom/huion-h950p.tablet` with that file. If the tablet then still shows up in `libwacom-list-local-devices`, we've won :)

But if the tablet is nicely packed away let's leave it there.  We're better off merging this and fixing it when someone complains who actually uses the tablet :) Thanks for the quick response though.

PR #571, opened 2023-06-26
Add generic groups for the generic styli

whot:
A readability improvement, nothing else. Poor souls that have to stare at a `.tablet` file are probably more comfortable reading `@generic-with-eraser` than `0xfffff;0xffffe`, doubly so since neither of those IDs actually exist or can be observed in the device data stream.


Pinglinux:
Excellent idea! Patchset looks good to me.

PR #579, opened 2023-07-11
data: add Huion Kamvas Pro 16 (2.5K) - huion-gt1620

filler-arc:
This should add support for Huion Kamvas Pro 16 (2.5K) huion-gt1620.
I copied kamvas-pro-13.svg since the button layout is the same.  

The only issue I couldn't figure out is that the display tablet shows up twice in the gnome-control-center panel, but other than that functionality seems to be in order.

Sorry if I've made a mess trying to understand pull requests and the like.

### DIFF
--- a/data/huion-gt1602.tablet
+++ b/data/huion-gt1602.tablet
@@ -1,0 +1,45 @@
+# Huion
+# Kamvas Pro 16 (2.5K)
+# GT1602
+#
+#
+#
+# Button Map:
+# (A=1, B=2, C=3, ...)
+#
+#    *-----------------------*
+#    |                       |
+#  A |                       |
+#  B |                       |
+#  C |                       |
+#  D |                       |
+#  E |        TABLET         |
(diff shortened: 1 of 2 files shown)


whot:
> The only issue I couldn't figure out is that the display tablet shows up twice in the gnome-control-center panel, but other than that functionality seems to be in order.

what's the output of `libwacom-list-local-devices` with this tablet file installed? I wonder if there are extra event nodes that get labelled.

also, please run the `sysinfo.sh` script from [wacom-hid-descriptors](https://github.com/linuxwacom/wacom-hid-descriptors) and file an issue in that repo, then link to the sysinfo in the `.tablet` file. The other data files have those links so check those for guidance, will make it easier in the future to get some info about the device if it doesn't work as expected.

whot:
gentle ping

whot:
closing, please reopen when the requested data has been provided, thanks

PR #588, opened 2023-08-31
data: add a new Keys/KeyCodes entry to the tablet files

### LINKED ISSUE #585: Cintiq Pro 13 Pad is not labeled as ID_INPUT_TABLET_PAD
Originally reported in [gnome-control-center#2019](https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/2019), cc @pinglinux

`tools/libwacom-update-db.py` treats devices without buttons as devices without pads and skips the hwdb entries for the `* Pad` suffix (see [here](https://github.com/linuxwacom/libwacom/blob/master/tools/libwacom-update-db.py#L79)). This results in the respective `Pad` device getting tagged as `ID_INPUT_TABLET` but not `ID_INPUT_TABLET_PAD`, making it appear as if there was more than one tablet connected.

This affects the Cintiq Pro 13, possibly others that don't have buttons but have a Pad event node.

This probably broke in f9cf43e431714240571c2c3b19ab6244f89dbb78, previously we ended up with a generic "* Pad" rule, thus tagging all devices ending in `Pad`.

whot:
This is parsed but not exported by libwacom itself, at least not until we know what our callers want.

It is used however to determine if a device has a pad and thus used to set the ID_INPUT_TABLET_PAD property through the hwdb. This fixes the issue with some devices without buttons (but with keys) to not get labelled as pads, e.g. the Cintiq Pro 13.

Fixes #585

Note: draft because I need to know the evdev keycodes for the cintiq before we can commit this, cc @Pinglinux

### DIFF
--- a/data/cintiq-pro-13.tablet
+++ b/data/cintiq-pro-13.tablet
@@ -43,3 +43,7 @@ Stylus=true
 Reversible=false
 Touch=true
 Ring=false
+
+[Keys]
+# first key is in-kernel display toggle
+KeyCodes=0;KEY_CONTROLPANEL;KEY_ONSCREEN_KEYBOARD;KEY_BUTTONCONFIG;SW_MUTE_DEVICE
--- a/data/wacom.example
+++ b/data/wacom.example
@@ -210,3 +210,10 @@ Touchstrip2=J
 # We assume the same number of modes for each of the touchstrips
 # if there is more than one
 StripsNumModes=4
+
+
+# Metadata about the keys on the tablet
+[Keys]
+# The evdev codes for the keys (if any) in order from left to right.
+# Keys that have a code of zero do not produce any events.
+KeyCodes=0;KEY_CONTROLPANEL;KEY_ONSCREEN_KEYBOARD;KEY_BUTTONCONFIG;SW_MUTE_DEVICE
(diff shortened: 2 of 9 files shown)


Pinglinux:
@whot, below are the details of the 5 touch keys. From left to right, on the top edge of the tablet, we see the keys as:
DISPLAY_TOGGLE (display off/on), KEY_CONTROLPANEL, KEY_ONSCREEN_KEYBOARD, KEY_BUTTONCONFIG, TOUCH on/off

DISPLAY_TOGGLE is managed in the kernel [1]. 
TOUCH on/off events are reported to userland through kernel SW_MUTE_DEVICE event [2]. SW_MUTE_DEVICE is also used to report physical touch switch events. You may want to ignore it or add special consideration for the event.
The other three keys are reported to userland as defined [3]. They are only used for soft keys, which are what you needed for this patch, I assume.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/hid/wacom_wac.c?h=v6.6-rc1&id=4082da80f46a6683439bb0357faadb18f0f5b2a6
[2] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/drivers/hid/wacom_wac.c?h=v6.6-rc1&id=d793ff81879a5747109f3106bd18e77662cb5187
[3] https://github.com/linuxwacom/input-wacom/commit/5d44f6f0ec74c68f57814ec7c0aff62fd9008ddd

whot:
@Pinglinux thanks for the info, I've updated the branch.

This turned out to get a bit more complicated, mostly because of the switch entry. I've filed #594 and #595 for the cleanups in the process.

One public API `libwacom_get_num_keys()`, the detailed API will come if we need it from a client somewhere.

PR #590, opened 2023-09-05
meson.build: correct license to HPND

whot:
Best explanation: 12-ish years ago when I started libwacom, I copied the license from some X11-related repository assuming it was MIT and went with that. Turns out it's actually the HPND-sell-variant: "Historical Permission Notice and Disclaimer - sell variant" https://spdx.org/licenses/HPND-sell-variant.html

No real effects since it's not really that much different to MIT afaict but let's make sure meson prints out the correct SPDX license identifier.

cc @jigpu  @Pinglinux

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project('libwacom', 'c',
 	version: '2.8.0',
-	license: 'MIT/Expat',
+	license: 'HPND',
 	default_options: [ 'c_std=gnu99', 'warning_level=2' ],
 	meson_version: '>= 0.51.0')
 


whot:
Note that the [Fedora License list](https://fedoraproject.org/wiki/Licensing:MIT?rd=Licensing/MIT#Old_Style_with_legal_disclaimer) refers this this as MIT "Old Style with legal disclaimer"

Pinglinux:
I was under the assumption that libwacom is on MIT license. If you say HPND is MIT "Old Style with legal disclaimer", it sounds "close" enough - we are developers anyway!

whot:
tbf, this is mostly legalese nit-picking, the HPND is a form of the MIT license but not *the* MIT license. SPDX has the various license texts as separate licenses but it's also becoming the reference for projects to point to the license they use. So we might as well go with that.

Effectively, nothing changes, neither for our users nor for our developers.

PR #600, opened 2023-10-27
data: remove erroneous dash from latest Cintiq Pro model names

skomra:


whot:
merged, thanks.

ftr patches like this feel free to just merge, the review isn't really necessary IMO

skomra:
> merged, thanks.
> 
> ftr patches like this feel free to just merge, the review isn't really necessary IMO

Great, thank you! noted!

PR #611, opened 2023-12-04
github: add a label commenter bot

whot:
This avoids us having to type the same thing out over and over again for the same (two most common) issues: sysinfo missing in a tablet file or a bug filed against libwacom that's not our bug.


whot:
merging, because we won't know if it works until it's on the master branch anyway

Issue #610, opened 2023-12-04
Fixing vendors that re-use USB IDs


whot:
Summary: we need some solution that allows us to match tablets that re-use the USB IDs. Huion, Gaomon and Ugee/XP-Pen seem to be the big offenders here and in some cases we can at least string match on the device name but e.g. #609 shows that's not always possible.

cc @JoseExposito - you've worked on some of these tablets in the kernel, is there anything we can export from the kernel to identify these tablets in userspace? Like stuffing an enum into `id_input->version`? Or some custom udev property? Or sysfs file?

TacticalCode:
I did some digging, with uclogic-tools (https://github.com/DIGImend/uclogic-tools) I can read details including the Firmware version off the attached device:


	$ sudo uclogic-probe 1 4 | uclogic-decode 
	    |           Manufacturer: GAOMON
	    |                Product: Gaomon Tablet
	 64 |        Params block #1: ???????????????????????????????????????????????????????????????????????????????????????????????????
	    |                          Max X: 8192
	    |                          Max Y: 357
	    |                   Max pressure: 2301
	    |                     Resolution: 2048
	    |
	 79 |         Internal model: ???????????????????????????????????????????????????????????????????????????????????????????????????
	 7b |         Buttons status: ???????????????????????????????????????????????????????????????????????????????????????????????????
	 c8 |        Params block #2: ??|?????
	    |                          Max X: 50800
	    |                          Max Y: 31750
	    |                   Max pressure: 8191
	    |                     Resolution: 5080
	    |
	 c9 |       Firmware version: OEM02_T19m_200114
	 ca |  Internal manufacturer: HUION Animation Technology Co.,ltd


Through the power of binwalk, strings and grep I found a `TabletCfg.dt` file in a download from the manufacturer that contains JSON data in `{ "FW_CODE": { "Properties":"values" } }` format for 129 firmware codes, 86 of which are unique non-null firmware -> ProductName pairs like:


	"OEM02_T19M": {
		"EnablePenKey0": false,
		"IsTouchHKey": false,
		"LED": false,
		"Langs": 131071,
		"MKeyConfig": null,
		"Multimedia": {
			"KeyCount": 0
		},
		"PenKeyCount": 3,
		"ProductName": "M106K PRO",
		"bMultimedia": false,
		"bOSD": false,
		"battery": false,
		"bluetooth": false,
		"dongle": false,
		"hasTitl": false,
		"hkeyNum": 12,
[...]

Issue #609, opened 2023-12-03
Add Gaomon M106K Support


TacticalCode:
Hello,

I'm using a Gaomon M106K pad, which is recognized as `Gaomon S620` due to identical hardware IDs.

The M106K is a "clone" of the Huion New 1060 Plus (see https://digimend.github.io/tablets/). The Gaomon M106K is already in wacom-hid-descriptors: https://github.com/linuxwacom/wacom-hid-descriptors/tree/master/Gaomon%20M106K
I've copied over the S620 data + Huion layout files to /etc/libwacom and modified the DeviceMatch and Buttons, now my pad is recognized correctly and I can configure all buttons in gnome settings.

My question is: Can we add support for the M106K without regression for the S620 support?

`dmesg` when I plug in the pad via USB:

	usb 1-2: new full-speed USB device number 4 using xhci_hcd
	usb 1-2: New USB device found, idVendor=256c, idProduct=006d, bcdDevice= 1.00
	usb 1-2: New USB device strings: Mfr=1, Product=2, SerialNumber=3
	usb 1-2: Product: Gaomon Tablet
	usb 1-2: Manufacturer: GAOMON
	hid-generic 0003:256C:006D.0006: hiddev96,hidraw1: USB HID v1.11 Device [GAOMON Gaomon Tablet] on usb-0000:00:14.0-2/input0
	input: GAOMON Gaomon Tablet Stylus as /devices/pci0000:00/0000:00:14.0/usb1/1-2/1-2:1.1/0003:256C:006D.0007/input/input27
	input: GAOMON Gaomon Tablet Keyboard as /devices/pci0000:00/0000:00:14.0/usb1/1-2/1-2:1.1/0003:256C:006D.0007/input/input28
	hid-generic 0003:256C:006D.0007: input,hidraw2: USB HID v1.11 Keyboard [GAOMON Gaomon Tablet] on usb-0000:00:14.0-2/input1
	input: GAOMON Gaomon Tablet Pen as /devices/pci0000:00/0000:00:14.0/usb1/1-2/1-2:1.0/0003:256C:006D.0006/input/input29
	input: GAOMON Gaomon Tablet Pad as /devices/pci0000:00/0000:00:14.0/usb1/1-2/1-2:1.0/0003:256C:006D.0006/input/input30
[...]

PR #614, opened 2023-12-04
Update gaomon-s620.tablet

FlashNoob98:
I updated the tablet firmware (through the windows tool) and now it displays as ' ID 256c:006f GAOMON Gaomon Tablet '
with lsusb, but the device is still not working on GNOME

### DIFF
--- a/data/gaomon-s620.tablet
+++ b/data/gaomon-s620.tablet
@@ -11,7 +11,7 @@
 [Device]
 Name=GAOMON S620
 ModelName=
-DeviceMatch=usb:256c:006d:GAOMON Gaomon Tablet Pen;usb:256c:006d:GAOMON Gaomon Tablet Pad;
+DeviceMatch=usb:256c:006f:GAOMON Gaomon Tablet;usb:256c:006d:GAOMON Gaomon Tablet Pen;usb:256c:006d:GAOMON Gaomon Tablet Pad;
 Class=Bamboo
 Width=6
 Height=4


FlashNoob98:
The device is now shown as S620 in Gnome settings but no input is given.
I got this error from libinput:
`event4  - GAOMON Gaomon Tablet Pen: libinput bug: missing tablet capabilities: resolution. Ignoring this device`
Otherwise, I solved the problem installing the driver from their website

`cat /lib/udev/rules.d/20-gaomon.rules 
# udev rules for GaoMonTablet supported devices
# This file is generated by GaoMonTablet driver, do not edit

KERNEL=="uhid",MODE:="0666",OPTIONS+="static_node=uhid"
KERNEL=="uinput",MODE:="0666",OPTIONS+="static_node=uinput"
SUBSYSTEMS=="usb",ATTRS{idVendor}=="256c",MODE:="0666"
SUBSYSTEM=="input", KERNEL=="event*",ATTRS{name}=="HUION 256C PEN STYLUS",MODE:="0644"
`

whot:
This means the kernel driver needs updating because the resolution may now be exported through some other bitfield that isn't yet parsed. In theory we could add an entry to `60-evdev.hwdb` to hardcode the resolution but we cannot do this for the Gaomon devices because they all share the same USB ID so we'd overwrite it for other devices too.

On that note: can you please run the sysinfo script from https://github.com/linuxwacom/wacom-hid-descriptors and attach the output in an issue there (see the readme in that repo).  Yesterday I looked for an S620 output and couldn't find any which prevented me to work on #610.

And second note: @FlashNoob98, meet @TacticalCode (and vice versa :). You two have an S620 and an M106K which re-use each other's USB ID, we're trying to figure out how to differ between the devices in userspace and it'd be great if you can help each other. See #609 and #610.

TacticalCode:
@whot My USB ID for the M106K is `256c:006d`, but I also haven't ever updated the firmware.

There also is a Gaomon "1080Pro" (reporting as "1060Pro") sysinfo with ID `256c:0064`: https://github.com/linuxwacom/wacom-hid-descriptors/issues/344
So I'm starting to think that maybe the S620 just had the wrong ID, or the manufacturer fixed the identical IDs with newer firmware?

@FlashNoob98 can you plug in your pad and run `libwacom-list-local-devices`?
If it's not listed there, you just need to override the current `DeviceMatch` string in the tablet definition:

	sudo -i
	mkdir /etc/libwacom
	sed -e 's/256c:006d/256c:006f/g' /usr/share/libwacom/gaomon-s620.tablet > /etc/libwacom/gaomon-s620.tablet
	libwacom-update-db

Re-attach the pad and run `libwacom-list-local-devices`, it should now show up and be configurable in gnome-settings.

Also, just for clarity, S620 is what you bought and not just what gnome-settings reports, right? Because right now most/all Gaomon pads are shown as "S620" in gnome-settings. That sysinfo would be really nice to have.

PR #617, opened 2023-12-17
data: Add Wacom HID-5362

olivgat:
Add laptop Lenovo Yoga 7 which uses Wacom HID 5362 Pen and Wacom HID 5362 Finger (vid: '0x056a', pid: '0x5362')

### DIFF
--- a/data/wacom-hid-5362.tablet
+++ b/data/wacom-hid-5362.tablet
@@ -1,0 +1,20 @@
+# Wacom
+# Wacom HID 5362
+#
+# sysinfo.oMdWcZJ918
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/325
+
+[Device]
+Name=Wacom HID 5362 Pen,Wacom HID 5362 Finger
+ModelName=
+Class=ISDV4
+DeviceMatch=i2c:056a:5362
+Width=14
+Height=9
+IntegratedIn=Display;System
+
+[Features]
+Stylus=true
+Reversible=false
+Touch=true
+


olivgat:
Thanks for the feedback whot.

Someone already submitted a report for strictly the same model as mine : https://github.com/linuxwacom/wacom-hid-descriptors/issues/325

Can I reference this issue in my tablet file ? Or do I have to create my own issue on hid-descriptors (with my own sysinfo tarball) ?

whot:
you can reference that file, we only care about having the data linked to the device for potential future debugging, not who collected it :)

data/wacom-hid-5362.tablet
```diff
@@ -0,0 +1,17 @@
...
+# Wacom HID 5362
+#
+# sysinfo.oMdWcZJ918.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/325
+
+[Device]
+Name=Wacom HID 5362 Pen,Wacom HID 5362 Finger
+ModelName=
+DeviceMatch=i2c:056a:5362
+IntegratedIn=Display;System
```
whot, line 11: Looks like `Width=` and `Height=` are missing here (that's the size it says on the box, in inches).

olivgat: I have added the dimensions according to the technical spec of the laptop.

whot:
Looks like the ` test-tablet-validity` test (part of `meson test -C builddir`) still fails, I can't immediately see the issue with the tablet file but the meson logs may have more.

olivgat:
> Looks like the ` test-tablet-validity` test (part of `meson test -C builddir`) still fails, I can't immediately see the issue with the tablet file but the meson logs may have more.

The culprit was this line in the log :
```
# Start of 3583 tests
not ok 339 /tablet/class/3583/056a:5362-Wacom HID 5362 Pen,Wacom HID 5362 Finger
```

I have added the Class parameter in my file and the test is now ok locally. I'm pushing my change to check if the workflow is ok too.

whot:
I just fixed the sysinfo check in https://github.com/linuxwacom/libwacom/pull/623, can you please rebase on top of that? Thanks.

Issue #602, opened 2023-11-07
Add support for the dual-tablet on the Lenovo Yoga Book 9i


whot:
See [gnome-control-center#2727](https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/2727), the Lenovo Yoga Book 9i has two built-in identical tablets currently listed as

```
- name: 'INGENIC Gadget Serial and keyboard Stylus'
  bus: 'usb'
  vid: '0x17ef'
  pid: '0x6161'
  nodes: 
  - /dev/input/event10: 'INGENIC Gadget Serial and keyboard Touchpad'
  - /dev/input/event9: 'INGENIC Gadget Serial and keyboard Stylus'
  - /dev/input/event8: 'INGENIC Gadget Serial and keyboard Stylus'
  - /dev/input/event7: 'INGENIC Gadget Serial and keyboard'
```

We may need some extra information to tell libwacom consumers that this is a dual-tablet device, though right now I'm not sure how that information would look like or how it'd be used.

whot:
cc @mistertwo

mistertwo:
Hey @whot,

Thanks for opening up an issue over here. One place I can see a "unified" enumeration of the screens would be in a drawing/art program so the tools/brush isn't switched between screens (assuming one is drawing across both).

Beyond that, it seems like most things would work with the screens each being addressed as their own entities, as libwacom and GSD are trying to do. Maybe have the screens handle separately for now, then look into the implementation of a DUAL_SCREEN class?

whot:
fwiw libwacom is just a text-file database with a C library wrapper. It won't control how things work but it could provide a hint like "this is a dual-screen device" which then causes g-c-c or something to search for another device with the same PID and map them accordingly. Or something :)

IOW this issue here is really just about how we can provide that hint, how exactly it'll be used will depend on the caller.

whot:
So, uhm, I'm staring at this again and I'm going to close this here.

For this to work we need something in the system to tell us what the difference is between these tablets (we talked about UNIQ iirc). Once we have something that can tell them apart we can figure out if we need something else in libwacom but right now I'm leaning towards "we don't" since the situation isn't that different to e.g. two Cintiqs plugged in.

Which means: I don't think libwacom needs anything here. We can re-open this later if we do though, so this is not a final decision.

mrojasampuero:
[...]

PR #621, opened 2023-12-23
adding .tablet file for Ingenic touch and pen display.

mistertwo:
This is in reference to issue #602 . Adds support for the Ingenic touch/pen screens on the Lenovo Yoga Book 9i.

Edited 2Jan23 to correct laptop model name.

### DIFF
--- a/data/ingenic-6161.tablet
+++ b/data/ingenic-6161.tablet
@@ -2,7 +2,7 @@
 
 [Device]
 Name=Ingenic Touch/Pen display
-ModelName=Gadget_Serial_and_keyboard
+ModelName=
 Class=Cintiq
 DeviceMatch=usb:17ef:6161
 Width=11
@@ -15,4 +15,3 @@ Stylus=true
 Reversible=false
 Touch=true
 Ring=false
-Buttons=0
--- a/data/ingenic-6161.tablet
+++ b/data/ingenic-6161.tablet
@@ -1,0 +1,18 @@
+# Ingenic Touch/Pen Display
+
+[Device]
+Name=Ingenic Touch/Pen display
+ModelName=Gadget_Serial_and_keyboard
+Class=Cintiq
+DeviceMatch=usb:17ef:6161
+Width=11
+Height=6
+# No pad buttons, so no layout
+IntegratedIn=Display;System
+
+[Features]
+Stylus=true
+Reversible=false
+Touch=true
+Ring=false
+Buttons=0
--- a/data/ingenic-6161.tablet
+++ b/data/ingenic-6161.tablet
@@ -1,4 +1,8 @@
 # Ingenic Touch/Pen Display
+# Integrated into the Lenovo Yoga Book 9i (13IRU8)
+# 
+# sysinfo.zpUboT0IUG.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/346
 
 [Device]
 Name=Ingenic Touch/Pen display
--- a/data/ingenic-6161.tablet
+++ b/data/ingenic-6161.tablet
@@ -7,7 +7,7 @@
 [Device]
 Name=Ingenic Touch/Pen display
 ModelName=
-Class=Cintiq
+Class=PenDisplay
 DeviceMatch=usb:17ef:6161
 Width=11
 Height=6


data/ingenic-6161.tablet
```diff
@@ -0,0 +1,18 @@
...
+Height=6
+# No pad buttons, so no layout
+IntegratedIn=Display;System
+
+[Features]
+Stylus=true
+Reversible=false
+Touch=true
+Ring=false
+Buttons=0
```
whot, line 18:
```suggestion
```

`Buttons=` was removed in libwacom 2.9 and is now implied by the content of an (optional) `[Buttons]` section.

data/ingenic-6161.tablet
```diff
@@ -0,0 +1,18 @@
+# Ingenic Touch/Pen Display
+
+[Device]
+Name=Ingenic Touch/Pen display
+ModelName=Gadget_Serial_and_keyboard
```
whot, line 5:
```suggestion
ModelName=
```
The model name is for cases where there's a marketing name shared across multiple models (Intuos Pro) but each one has a specific model name too. I think in this case we can just skip this, this looks like a template name that doesn't really specify this particular model.

mistertwo: Change made in branch.

whot: thanks, I'll still need the sysinfo link please

mistertwo: Uploaded here: https://github.com/linuxwacom/wacom-hid-descriptors/issues/346

whot:
I just fixed the sysinfo check in #623, can you please rebase on top of that? Thanks.

whot:
hmm, looks like this fails `meson test -C builddir` because the class is `Cintiq` and for that we require actual styli to be specified. From the [gnome-control-center issue](https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/2727) I can see your device has a serial number so you would need something like `Styli=@isdv4-aes`.

The stylus tool ID is the `ABS_MISC` value in `libinput record` (as opposed to `MSC_SERIAL` which is the unique pen serial) - that should (hopefully) have an entry in `data/libwacom.stylus` - then you specify the whatever group that stylus belongs to. e.g. if the id is `0x81b` then the line would be `Styli=@intuos3`.

mistertwo:
> The stylus tool ID is the `ABS_MISC` value in `libinput record` (as opposed to `MSC_SERIAL` which is the unique pen serial) - that should (hopefully) have an entry in `data/libwacom.stylus` - then you specify the whatever group that stylus belongs to. e.g. if the id is `0x81b` then the line would be `Styli=@intuos3`.

I've checked into the ABS_MISC value via **libinput record**, and it appears to be something that determines distance of the pen from the display. It's incrementing by ~35 for each input "frame" and increases the increment value as the pen is pulled away from the screen.

```
    libinput:
    - {time: 0.106812, type: TABLET_TOOL_AXIS, tip: up, point: [7.82, 43.09], tilt: [25.70, 16.68], pressure: 0.00}
  - evdev:
    - [  0, 111793,   3,   0,   27760] # EV_ABS / ABS_X                 27760 (+12)
    - [  0, 111793,   3,   1,   13409] # EV_ABS / ABS_Y                 13409 (-12)
    - [  0, 111793,   3,  26,    2555] # EV_ABS / ABS_TILT_X             2555 (-5)
    - [  0, 111793,   3,  27,    1682] # EV_ABS / ABS_TILT_Y             1682 (+4)
    - [  0, 111793,   3,  40,   21824] # EV_ABS / ABS_MISC              21824 (+38)
    - [  0, 111793,   0,   0,       0] # ------------ SYN_REPORT (0) ---------- +5ms
    libinput:
    - {time: 0.111793, type: TABLET_TOOL_AXIS, tip: up, point: [7.75, 43.19], tilt: [25.65, 16.74], pressure: 0.00}
  - evdev:
    - [  0, 113811,   3,   0,   27776] # EV_ABS / ABS_X                 27776 (+16)
    - [  0, 113811,   3,   1,   13396] # EV_ABS / ABS_Y                 13396 (-13)
    - [  0, 113811,   3,  26,    2551] # EV_ABS / ABS_TILT_X             2551 (-4)
    - [  0, 113811,   3,  27,    1685] # EV_ABS / ABS_TILT_Y             1685 (+3)
    - [  0, 113811,   3,  40,   21862] # EV_ABS / ABS_MISC              21862 (+38)
    - [  0, 113811,   0,   0,       0] # ------------ SYN_REPORT (0) ---------- +2ms
```

Would the **PenDisplay** class fit this situation better? Also, if I'm not reading the **libinput record** reports right, please let me know.

data/ingenic-6161.tablet
```diff
@@ -0,0 +1,21 @@
+# Ingenic Touch/Pen Display
+# Integrated into the Lenovo Yoga Book 9i (13IRU8)
+# 
+# sysinfo.zpUboT0IUG.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/346
+
+[Device]
+Name=Ingenic Touch/Pen display
+ModelName=
+Class=Cintiq
```
whot, line 10:
```suggestion
Class=PenDisplay
```

whot:
The `ABS_MISC` tool id is constant for a pen and never changes (on wacom devices anyway). So if you get something else in `ABS_MISC` that's a sign the kernel driver is not doing the right thing. 

> It's incrementing by ~35 for each input "frame" and increases the increment value as the pen is pulled away from the screen.

That sounds like maybe two separate things in the HID report are combined into one axis?
cc @JoseExposito, @bentiss in case they have time, but please attach a `hid-recorder` output here too (https://gitlab.freedesktop.org/libevdev/hid-tools/), that should tell us more.

> Would the PenDisplay class fit this situation better

Yeah, let's try that, we can change it later if we end up with serial numbers. The `Class` isn't actually used by anything that I know of, so it's just a matter of getting past the test suite anyway here.

mistertwo:
@bentiss the delta between frames decreases as the pen tip gets closer to the screen. The actual value continues to increment until it hits FFFF, then it recycles.

For completeness, here's a yaml of output from libinput record for the pen tip on the upper screen. Movement contents:
- slow movement of the pen to the screen
- two touches of the tip to the screen
- slow pullback from the screen.

[ingenic_stylus.zip](https://github.com/linuxwacom/libwacom/files/14050843/ingenic_stylus.zip)

PR #625, opened 2024-01-14
add support for xp-pen artist 10s table

hooger:
I have som trouble with this descriptor, please help me with it.

- I had to use the string that is returned by `libinput list-devices` for kwin to recognize the descriptor. However, this is quite weird as the tablet is an XP-Pen Artist 10s, and it is not obvious what an UGEE device is. That would be nice to overwrite, yet the ModelName key has made no difference. Am I doing something wrong, or it is a problem with kwin/ksystem settings?
- In the example table file there is something about a system identifier generated by wacom-hid-descriptor. Howevert, that project seems to be about devices that are drawing computers, while this device is an external, usb connected drawing tablet. Is it alright that I omitted these comment lines or should I extend the .table file somehow?

### DIFF
--- a/data/xp-pen-artist10s.tablet
+++ b/data/xp-pen-artist10s.tablet
@@ -4,12 +4,12 @@
 
 [Device]
 Name=UGEE 10.1 Tablet Monitor
-ModelName=XP-Pen Artist 10S (1st gen)
+ModelName=
 DeviceMatch=usb:5543:004a:UGEE 10.1 Tablet Monitor;usb:5543:004a:UGEE 10.1 Tablet Monitor Pad;usb:5543:004a:UGEE 10.1 Tablet Monitor Mouse
 PairedIDs=
 Class=Bamboo
-Width=8.5
-Height=5.3
+Width=9
+Height=5
 IntegratedIn=
 Layout=xp-pen-artist10s.svg
 Styli=@generic-no-eraser;
--- a/data/xp-pen-artist10s.tablet
+++ b/data/xp-pen-artist10s.tablet
@@ -1,6 +1,8 @@
 # XP-Pen
 # Artist 10S (1st generation)
 #
+# sysinfo.2kd91mftwQ
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/354
 
 [Device]
 Name=UGEE 10.1 Tablet Monitor
--- a/data/xp-pen-artist10s.tablet
+++ b/data/xp-pen-artist10s.tablet
@@ -23,7 +23,6 @@ Touch=false
 TouchSwitch=false
 Ring=false
 Ring2=false
-NumStrips=1
 
 [Buttons]
 Left=A;B;C;D;E;F
--- a/data/xp-pen-artist10s.tablet
+++ b/data/xp-pen-artist10s.tablet
@@ -1,0 +1,28 @@
+# XP-Pen
+# Artist 10S (1st generation)
+#
+
+[Device]
+Name=UGEE 10.1 Tablet Monitor
+ModelName=XP-Pen Artist 10S (1st gen)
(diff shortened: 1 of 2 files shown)


whot:
>  is not obvious what an UGEE device is

UGEE is afaik the parent company of XP-Pen, so some tablets get rebranded, others get rebranded by sticking them in a XP Pen box without changing the firmware to remove the UGEE references.

It's fine to list the Name as what it says on the box and have the `DeviceMatch` match on UGEE. libwacom only provides this as a hint, if kwin prefers to display the kernel name then there's nothing you can do (in libwacom) to change that.


> Is it alright that I omitted these comment lines or should I extend the .table file somehow?

Please extend it, e.g. all the Intuos tablets already in libwacom are external devices like the one you have. Until a few years ago, those were the majority before all the built-in ones started getting more numerous.

data/xp-pen-artist10s.tablet
```diff
@@ -0,0 +1,28 @@
...
+#
+
+[Device]
+Name=UGEE 10.1 Tablet Monitor
+ModelName=XP-Pen Artist 10S (1st gen)
+DeviceMatch=usb:5543:004a:UGEE 10.1 Tablet Monitor;usb:5543:004a:UGEE 10.1 Tablet Monitor Pad;usb:5543:004a:UGEE 10.1 Tablet Monitor Mouse
+PairedIDs=
+Class=Bamboo
+Width=8.5
+Height=5.3
```
whot, line 12: These need to be integers, so please just round them to whatever is closest. It doesn't *really* matter anyway (not sure anything actually uses those)

data/xp-pen-artist10s.tablet
```diff
@@ -0,0 +1,28 @@
+# XP-Pen
+# Artist 10S (1st generation)
+#
+
+[Device]
+Name=UGEE 10.1 Tablet Monitor
+ModelName=XP-Pen Artist 10S (1st gen)
```
whot, line 7:
```suggestion
ModelName=
```
ModelName doesn't really apply to most XP pen/UGEE tablets, more to Wacom where the "Wacom Intuos blah" has a model name of "ABC-123".

data/xp-pen-artist10s.tablet
hooger: fixed

data/xp-pen-artist10s.tablet
hooger: fixed

hooger:
> Please extend it, e.g. all the Intuos tablets already in libwacom are external devices like the one you have. Until a few years ago, those were the majority before all the built-in ones started getting more numerous.

I think I added all the necessary information in https://github.com/linuxwacom/wacom-hid-descriptors/issues/354 Can you take a look again, please?

data/xp-pen-artist10s.tablet
```diff
@@ -0,0 +1,30 @@
...
+Styli=@generic-no-eraser;
+
+[Features]
+Stylus=true
+Reversible=true
+Touch=false
+TouchSwitch=false
+Ring=false
+Ring2=false
+NumStrips=1
```
whot, line 26: `Bail out! ERROR:../test/test-tablet-svg-validity.c:136:check_touchstrip: assertion failed: (node != NULL)`

CI is complaining because the svg doesn't have a touch strip defined. Looks like the `data/xp-pen-artist12.tablet` file has one strip defined too so you can probably peek at that svg for guidance.

hooger: Ah, I see. Does this NumStrip means touch strip? This tablet (10s) has 6 buttons only, and they are sending event codes BTN_0 ... BTN_5 that's why I thought this is a NumStrip. Should I just remove it from the .tablet file then (NumStrip=0 is the default) ? See my commit.

whot: yep, `NumStrips` defines the number of touch strips. If yours doesn't have any of those then setting it to zero (or leaving it out) is enough.

whot:
Merged, thanks!

PR #628, opened 2024-01-24
Update Dell Active Pen

Doomsdayrs:
The Dell Active Pen is misconfigured in `libwacom.stylus`.

Firstly, The pen has 2 buttons.

Secondly "AES" pen does not properly illustrate what the pen is to the user.

### DIFF
--- a/data/libwacom.stylus
+++ b/data/libwacom.stylus
@@ -159,7 +159,7 @@ Type=Mobile
 # Dell   ; VID_NONE     | 0xf142541e | BAT_SWAP (Dell PN5122W)
 Name=Dell Active Pen
 Group=isdv4-aes
-Buttons=2
+Buttons=1
 EraserType=Button
 Axes=Pressure
 Type=Mobile
--- a/data/libwacom.stylus
+++ b/data/libwacom.stylus
@@ -156,9 +156,10 @@ Type=Mobile
 
 [0x631]
 # Dell   ; VID_NONE     | 0x0000 | BAT_SWAP | LONGPRESS (Dell PN557W)
-Name=AES Pen
+# Dell   ; VID_NONE     | 0xf142541e | BAT_SWAP (Dell PN5122W)
+Name=Dell Active Pen
 Group=isdv4-aes
-Buttons=1
+Buttons=2
 EraserType=Button
 Axes=Pressure
 Type=Mobile


whot:
> Firstly, The pen has 2 buttons.

2 *actual* buttons or 1 button + an eraser button? The latter doesn't count as button (iirc) since we can't re-map it, it's hardcoded for some specific behaviour that requires heuristics to undo and we'll likely never implement those. So even if the pen has 2 physical buttons, from libwacom's POV there's only one that is a real button.

Doomsdayrs:
> > Firstly, The pen has 2 buttons.
> 
> 2 _actual_ buttons or 1 button + an eraser button? The latter doesn't count as button (iirc) since we can't re-map it, it's hardcoded for some specific behaviour that requires heuristics to undo and we'll likely never implement those. So even if the pen has 2 physical buttons, from libwacom's POV there's only one that is a real button.

Thanks for the explanation! 

I was confused why I only saw one button listed in the GNOME Settings UI, I shall take it there to help explain that to the user.

---

I hope the name change is okay?

---

Furthermore, I notice that the SVG defaults to this "basic" pen. Is there any way I can provide an SVG for my pen, I did not see documentation about it.

image

whot:
for stylistic reasons [1] the svgs for the syli are part of the gnome-control-center, so you'd need to add the pen there and then check the code on how the SVGs are being picked based on the current tool.

The name change is fine, let's hope this id is only used by Dell (if not, we'll have to change it back).


[1] the tablet SVGs are neutral-ish, the svgs for the pens were more specific to the gnome style, that's what I vaguely remember.

data/libwacom.stylus
```diff
@@ -156,9 +156,10 @@ Type=Mobile
 
 [0x631]
 # Dell   ; VID_NONE     | 0x0000 | BAT_SWAP | LONGPRESS (Dell PN557W)
-Name=AES Pen
+# Dell   ; VID_NONE     | 0xf142541e | BAT_SWAP (Dell PN5122W)
+Name=Dell Active Pen
 Group=isdv4-aes
-Buttons=1
+Buttons=2
```
whot, line 162:
```suggestion
Buttons=1
```

whot:
Realised I can just fix the buttons line here, so let's do that. Merged, thanks!

PR #634, opened 2024-02-05
data: Intuos Pro 2 is compatible with Intuos 4 styli

whot:
The Windows driver at least blocks those but they work just fine. Add them here so we can look them up.

cc @jigpu, @Pinglinux, @Joshua-Dickens 

Note: this is required for [gnome-control-center!2244](https://gitlab.gnome.org/GNOME/gnome-control-center/-/merge_requests/2244) for the tablet to work with the old pens.


Pinglinux (approved):
Works for me. Thank you Peter!

PR #640, opened 2024-02-19
data: add ELAN 9008 and 9009 (zenbook duo UX8406MA)

alesya-h:
Adding both touch panels for the new dualscreen Asus Zenbook Duo


whot:
merged, thanks

PR #644, opened 2024-03-11
Add lenovo IdeaPad Duet 3 10IGL5 (elan-2bd6)

knezi:
The tablet works, but not sure about the stylus.

`tools/show-stylus.py` says it's only capable of generic stylus.

In Gnome settings I can see the stylus, but changing the button mappings doesn't do anything.

`libinput record` shows some event when pressing stylus buttons, but look random to me. When pressing buttons, sometimes I get a regular key press, sometimes nothing and sometimes eraser/rubber.  Also when using the stylus buttons, they every time produce different event.

Is there anything I can do about this?

### DIFF
--- a/data/elan-2bd6.tablet
+++ b/data/elan-2bd6.tablet
@@ -1,0 +1,16 @@
+# ELAN touchscreen/pen sensor present in the IdeaPad Duet 3 10IGL5
+#
+# sysinfo.Al8cFbaepN
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/212
+
+[Device]
+Name=Elan 2BD6
+ModelName=
+DeviceMatch=i2c:04f3:2bd6
+Class=ISDV4
+IntegratedIn=Display;System
+Styli=@generic-with-eraser
+
+[Features]
+Stylus=true
+Touch=true


whot:
> In Gnome settings I can see the stylus, but changing the button mappings doesn't do anything.

That'd be an issue in the gnome-control-center/mutter then, libwacom just provides a static description of the device/stylus.

> libinput record shows some event when pressing stylus buttons, but look random to me.

You may have a stylus with an eraser button? In that case the button simulates (in firmware) taking the pen out of proximity, flipping it around to the eraser end and bringing it into proximity again.

Some styli also have a (Bluetooth BLE) button at the end which triggers special events that may be mapped to a keystroke by default.

attaching the libinput record output here can tell for sure (attach, not copy/paste please)

knezi:
Yes, you're right. The lower button indeed is erasers and this touch and lift simulation happens. I've updated the PR to stylus with eraser. Xournal and Gnome still don't work as expected and have one button too many in the config, but that's not an issue with libwacom as `libinput record` produces correct events, so I think it's ready for merging.

whot:
Looks like the extra button is an issue with libwacom - the generic stylus has 2 buttons and that's probably where the extra button comes from. I think we could work around this but... meh? :)

PR #648, opened 2024-03-22
Add support for firmware versions

jexposit:
Patches from Peter (https://github.com/linuxwacom/libwacom/pull/620) plus changes from my side.

Requires https://lore.kernel.org/linux-input/20240322100210.107152-1-jose.exposito89@gmail.com/T/ or https://github.com/JoseExposito/digimend-kernel-drivers/tree/sysfs-fw-name.

Fixes https://github.com/linuxwacom/libwacom/issues/610

cc @whot

### DIFF
--- a/data/wacom.example
+++ b/data/wacom.example
@@ -76,11 +76,11 @@ ModelName=
 # so it has exactly 4 digits. hex letters must be lowercase.
 #
 # Multiple matches are joined by a semicolon
-DeviceMatch=usb:056a:00bc;bluetooth:056a:00bc;
+DeviceMatch=usb|056a|00bc;bluetooth|056a|00bc;
 # If the vendor/product ID is not sufficient to distinguish the name
 # (common for HUION, GAOMON, UCLogic, XP-Pen), use the device name
 # as fourth element in the match string.
-# DeviceMatch=usb:1234:abcd:SomeDevice Name;
+# DeviceMatch=usb|1234|abcd:SomeDevice Name;
 
 
 # Paired PID includes the match line of any device that share the same
@@ -88,7 +88,7 @@ DeviceMatch=usb:056a:00bc;bluetooth:056a:00bc;
 # device on the 24HDT). The format of the match line is identical to
 # DeviceMatch but only one value is permitted.
 # Note: the PIDs listed may not be libwacom devices themselves.
-PairedIDs=usb:056a:0335
+PairedIDs=usb|056a|0335
 
 # Class of the tablet. Valid classes include Intuos3, Intuos4, Graphire, Bamboo, Cintiq
 # If unsure, or not applicable (the tablet isn't stand-alone for example).
--- a/data/wacom.example
+++ b/data/wacom.example
@@ -81,7 +81,11 @@ DeviceMatch=usb|056a|00bc;bluetooth|056a|00bc;
 # (common for HUION, GAOMON, UCLogic, XP-Pen), use the device name
 # as fourth element in the match string.
 # DeviceMatch=usb|1234|abcd:SomeDevice Name;
-
+#
+# Optionally after the device name (which may be empty) the
+# the remainder of that substring is the uniq ID to match against.
+# DeviceMatch=usb|1234|abcd||ABC_1234;
+# DeviceMatch=usb|1234|abcd|SomeDevice Name|ABC_1234;
 
 # Paired PID includes the match line of any device that share the same
 # physical device but has different product or vendor ids (e.g. the touch
--- a/libwacom/libwacom.c
+++ b/libwacom/libwacom.c
@@ -212,11 +212,45 @@ client_query_by_subsystem_and_device_file (GUdevClient *client,
 	return ret;
 }
 
+static char *
+get_device_uniq(GUdevDevice *device)
+{
+	GRegex *regex;
+	GMatchInfo *match_info;
+	gchar *uniq = g_strdup (g_udev_device_get_sysfs_attr (device, "uniq"));
+
+	if (!uniq)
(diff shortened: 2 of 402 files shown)


data/wacom.example
```diff
@@ -74,14 +74,25 @@ ModelName=
...
+#
+# Multiple matches are joined by a semicolon
+DeviceMatch=usb|056a|00bc;bluetooth|056a|00bc;
+# If the vendor/product ID is not sufficient to distinguish the name
+# (common for HUION, GAOMON, UCLogic, XP-Pen), use the device name
+# as fourth element in the match string.
+# DeviceMatch=usb|1234|abcd:SomeDevice Name;
+#
+# Optionally after the device name (which may be empty) the
+# the remainer of that substring is the firmware version to match against.
```
whot, line 86:
```suggestion
# the remainder of that substring is the firmware version to match against.
```

libwacom/libwacom.c
```diff
@@ -212,11 +212,43 @@ client_query_by_subsystem_and_device_file (GUdevClient *client,
...
+		return NULL;
+
+	/* The UCLogic kernel driver returns firmware names with form
+	 * <vendor>_<model>_<version>. Remove the version from `fw` to avoid
+	 * mismatches on firmware updates. */
+	regex = g_regex_new ("(.*_.*)_.*$",
+				G_REGEX_DEFAULT,
+				G_REGEX_MATCH_DEFAULT,
+				NULL);
+	g_regex_match (regex, fw, 0, &match_info);
```
whot, line 232: Note to self: this may become an issue if we ever get a fw exported that has a different underscorify pattern.

whot (approved):
One typo found, would be great if you could squash that in. Otherwise this LGTM, we can merge this once upstream accepts your kernel patch. Can you ping me when that happens? Thanks

jexposit:
@whot the [kernel patch has been merged](https://lore.kernel.org/linux-input/20240322100210.107152-1-jose.exposito89@gmail.com/T/#e66606679b7c120427b34cecbb5fe9ff4d921c1d5).

In the end, instead of creating a custom sysfs property, we ended up using `hid->uniq`. Since this a generic attribute all devices has, I wonder if it makes sense to refactor a little bit the series to use `uniq` instead of `fw` as variable names and update the documentation.

Something else I have in my to-do list is displaying that information with `libwacom-list-local-devices` so user can actually discover it.

whot:
> I wonder if it makes sense to refactor a little bit the series to use uniq instead of fw as variable names and update the documentation.

yes, definitely. This also leaves room for a future firmware attribute if uniq is no longer sufficient. 

> Something else I have in my to-do list is displaying that information with libwacom-list-local-devices so user can actually discover it.

I think `--format=datafile` should work automatically, the yaml format will need an extra entry though.

jexposit:
> > I wonder if it makes sense to refactor a little bit the series to use uniq instead of fw as variable names and update the documentation.
> 
> yes, definitely. This also leaves room for a future firmware attribute if uniq is no longer sufficient.

Refactored the last patch to use uniq instead of fw

> > Something else I have in my to-do list is displaying that information with libwacom-list-local-devices so user can actually discover it.
> 
> I think `--format=datafile` should work automatically, the yaml format will need an extra entry though.

Also added an extra entry to `tools/list-devices.c`

libwacom/libwacom.c
```diff
@@ -817,6 +876,7 @@ static void print_integrated_flags_for_device (int fd, const WacomDevice *device
 static void print_match(int fd, const WacomMatch *match)
 {
 	const char  *name       = libwacom_match_get_name(match);
+	const char  *uniq         = libwacom_match_get_uniq(match);
```
whot, line 879:
```suggestion
	const char  *uniq        = libwacom_match_get_uniq(match);
```

whot (approved):
one indentation error, rest looks good, thanks!

libwacom/libwacom.sym
```diff
@@ -72,3 +72,7 @@ local:
 LIBWACOM_2.9 {
     libwacom_get_num_keys;
 } LIBWACOM_2.0;
+
+LIBWACOM_2.11 {
```
whot, line 76:
```suggestion
LIBWACOM_2.12 {
```

(because I just released 2.11, sorry)

jexposit:
Resolved and rebased. Thanks for the review!

whot:
`../libwacom/libwacom.c:234:33: error: ‘G_REGEX_DEFAULT’ undeclared (first use in this function); did you mean ‘G_REGEX_DOTALL’?`

Introduced in 2.74 and that's too new for Ubuntu 22.04. Looking at glib commit `2.73.1-26-g879b9cd669f0` we should be fine with an
```
#ifndef G_REGEX_DEFAULT
#define G_REGEX_DEFAULT 0
#endif
```

jexposit:
Ops, missed the CI emails. I replaced `compile_options` and `match_options` with 0 as in the docs:
https://docs.gtk.org/glib/ctor.Regex.new.html

jexposit:
Is the FreeBSD failure expected?

whot:
> Is the FreeBSD failure expected?

yeah, it's very unreliable and fails often (all the time?) but I'm ETIME on debugging that. cc @jbeich

whot:
I think we're good here, let's merge this! thanks for all the effort

jbeich:
> it's very unreliable

Try applying https://github.com/labwc/labwc/commit/00ec29c834b7. I suspect `v0` isn't maintained anymore or emulating on macOS is still unstable.

whot:
@jbeich thanks, that was it, fixed in #662

PR #653, opened 2024-04-04
add wacom hid 52EB used in thinkpad l13 yoga

teberhardt:


data/wacom-hid-52EB-pen.tablet
```diff
@@ -0,0 +1,22 @@
+# Wacom
+# Wacom HID 52EB Pen
+# WACF2200
+#
+# sysinfo.0VeQaUVnQ1
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/369
+
+[Device]
+Name=Wacom HID 52EB Pen, Wacom HID 52EB Finger
```
whot, line 9:
```suggestion
Name=Wacom HID 52EB Pen
```
The finger device is the touch device and doesn't have any effect on libwacom so better to leave it out

data/wacom-hid-52EB-pen.tablet
```diff
@@ -0,0 +1,22 @@
...
+ModelName=WACF2200
+Class=ISDV4
+DeviceMatch=i2c:056a:52eb;
+# Width=11
+# Height=6
```
whot, line 14: These two are required, please add whatever the closest w/h is in inches, rounded.

whot (changes requested):
Minor nitpicks but this is mostly ready, thanks

data/wacom-hid-52EB-pen.tablet
teberhardt: Thanks for the fast response, fixed the nitpicks. 
13.3" in 16:10 should be W=11 and H=7.

whot: >  13.3" in 16:10 should be W=11 and H=7.

heh, can't you just measure the thing? It's in front of you, right? :)
Anyway, afaict no-one really looks at these numbers anyway, so I'm happy with 11/7.

whot:
Merged, thanks

PR #657, opened 2024-04-18
data: Add Gaomon S620 firmware match

jexposit:
This tablet has the same name as Gaomon M106K [1], making it impossible to differentiate between them without matching by their firmware name.

Add an extra match including the firmware to facilitate matching the Gaomon M106K.

[1] https://github.com/linuxwacom/libwacom/issues/609

### DIFF
--- a/data/gaomon-s620.tablet
+++ b/data/gaomon-s620.tablet
@@ -11,7 +11,7 @@
 [Device]
 Name=GAOMON S620
 ModelName=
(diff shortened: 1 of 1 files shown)


whot:
Sorry, the summary confused me a bit: this is the firmware for the S620 here right? And the M106K will go into a separate file?

whot:
Ah, nevermind, the comment in #609 clears this up

whot:
merged, thanks

PR #658, opened 2024-04-19
test: add some tracing to the svg validity test

whot:
It's a bit hard to figure out from the CI logs what failed because they may be truncated. And for users not intimitely familiar with the test suite it's not that simple either so let's put in a bunch of trace stataments that should at least point to where things go wrong.

cc @AesaraB

### DIFF
--- a/test/test-tablet-svg-validity.c
+++ b/test/test-tablet-svg-validity.c
@@ -36,6 +36,9 @@
 #include <glib.h>
 #include "libwacom.h"
 
+#define trace(fmt, ...) \
+	fprintf(stderr, "%s: checking " fmt "\n", __func__,  __VA_ARGS__)
+
 static xmlNodePtr
 verify_has_sub (xmlNodePtr cur, char *sub)
 {
@@ -79,11 +82,12 @@ class_found (gchar **classes, gchar *value)
 static void
 verify_has_class (xmlNodePtr cur, const gchar *expected)
 {
-
 	xmlChar *prop;
 	gchar  **classes_present;
 	gchar  **classes_expected;
 	gchar  **ptr;
+
+	trace("%s", expected);
 
 	prop = xmlGetProp (cur, (xmlChar *) "class");
 	g_assert (prop != NULL);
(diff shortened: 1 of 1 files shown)


AesaraB:
Mildly tangential, but I made a fairly rudimentary bash script to debug things.

```bash
#!/usr/bin/env bash

FILENAME=$1

echo "Checking for invalid class/id names"
grep -P "(class|id)=[\"'].*[\"']" $FILENAME | grep -vP '(class="[A-Z] (Button|Leader|Label)|id="(Button|Leader|Label)[A-Z])"'

echo "Checking for duplicate ids"
grep -P "id=[\"'](Button|Leader|Label)[A-Z][\"']" $FILENAME | sort | uniq -cd

echo "Checking for duplicate classes"
grep -P "class=[\"'][A-Z] (Button|Leader|Label)[\"']" $FILENAME | sort | uniq -cd
```

whot:
honestly, I think the whole svg validity test should be rewritten in python + pytest - it'll be simpler and provide better debugging options than the current one (which is now 12y old...)

AesaraB:
> honestly, I think the whole svg validity test should be rewritten in python + pytest - it'll be simpler and provide better debugging options than the current one (which is now 12y old...)

I'm unfamiliar with low level programming, but I just heard some terms and ideas that I'm relatively familiar with.

I have some housekeeping to do with other projects and my personal life, so if I get those sorted before someone else commits their time to this, I think I should be able to write that code.

whot:
heh, I don't think in this case it's really low-level programming - you can parse the data files in python (`ConfigParser` will do it) and then for each svg file check that the expected bits exist.

We already have a python test to check the data files exist (`test_svg_exists()`), it would be a similar test to also check that for each Button `A` we have an svg element with `id="Button A"` that also has `class="A Button"`. Then repeat for rings/strips/etc.

Definitely a good one to get started with, doubly so if you want to learn python :)

whot:
Going to merge this since @JoseExposito also ran into the issue of the svg checks not being particularly useful with their error messages. The output is a bit spammy and I've changed to to stderr now so it doesn't conflict with the TAP output on stdout but it's still better than a comment-less `abort()`

whot:
> honestly, I think the whole svg validity test should be rewritten in python + pytest - it'll be simpler and provide better debugging options than the current one (which is now 12y old...)

This is now part of #677

PR #659, opened 2024-04-21
HUION: Add auto generated description files

JoseExposito:
This PR adds a bunch of HUION tablet description files and SVGs generated with this tool:
https://github.com/JoseExposito/uclogic-libwacom-tablet-descriptions

While incomplete, basic information like the number of buttons, number of touch strips and the device class (bamboo/cintiq) is correct.


data/huion-gc610-710.tablet
```diff
@@ -0,0 +1,26 @@
...
+# If you own this tablet, please improve it.
+
+[Device]
+Name=Huion GC610/710
+ModelName=
+Class=Bamboo
+DeviceMatch=usb|256c|006e||HUION_T166;usb|256c|006d||HUION_T166;
+Width=
+Height=
+Layout=
```
whot, line 12: See #661 and #660, these should get over this test case failure. Not super-happy about #661 (removing the need for a layout file) but if it gets us more tablet support it's still a net win.

whot (approved):
This LGTM and is obviously a massive boon in tablet files, so thanks heaps for that.

I'm wondering what the overlap is between these new devices and the ones we already have. IOW I think some of these new devices were already matched by existing files and may break/change now but then again that should bring the users out of the woodwork to fix the layout files :)

JoseExposito:
> I'm wondering what the overlap is between these new devices and the ones we already have. IOW I think some of these new devices were already matched by existing files and may break/change now but then again that should bring the users out of the woodwork to fix the layout files :)

I deleted manually the description files already present in libwacom to avoid clashes. Have I override any accidentally?

data/huion-gc610-710.tablet
JoseExposito: If an empty `Layout=` entry causes issues, I can delete it. The files are generated automatically, so it'd be a matter of dropping a line of code.

About #661, I added [a comment](https://github.com/linuxwacom/libwacom/pull/661#pullrequestreview-2014070578) suggesting a possible ~solution~ enhancement.

JoseExposito: Converted to draft as there might be a way to auto generate SVGs. I'm not sure about `Width`/`Height` yet... But I can always google 60 tablet sizes :sleepy:

whot: I'd say just add 9x6 and a comment and we're good. *Nothing* reads those values but for backwards compat (and in case someone will want them in the future) we still need to set them. If they're wrong, no-one's going to notice for the foreseeable future.

whot:
> I deleted manually the description files already present in libwacom to avoid clashes. Have I override any accidentally?

I don't see any deleted files in this MR so I'm not sure what you meant there? My comment was related to: huion re-uses the USB IDs and the names so we have `.tablet` files that match tablets they weren't intended to. With this MR this *should* be fixed but it may cause a change in a previously overmatched tablet. Again, it may happen but I'm not sure we need to worry about it...

whot:
To illustrate my comment above: the only `.tablet` file that is overwritten by the autogenerated results in https://github.com/JoseExposito/uclogic-libwacom-tablet-descriptions/pull/1 is the `HS611` and that shows the issue quite nicely:

```diff
- DeviceMatch=usb|256c|006d|HUION Huion Tablet_HS611 Pad;usb|256c|006d|HUION Huion Tablet_HS611 Touch Strip;usb|256c|006d|HUION Huion Tablet_HS611 Pen
+ DeviceMatch=usb|256c|006e||HUION_T19c;usb|256c|006d||HUION_T19c;
```

This device would indeed break if we were to replace the tablet file - anyone with older digimend patches won't get the uniq set and because we have no fallback match we won't match this device anymore.

This means we need to add the autogenerated patches *in addition* to the existing ones until we can reasonably expect users to carry the updated kernel patches. Our match system goes name+uniq first, then uniq, then name so for tablets where we have both uniq and old-style name match defined we should match correctly either way.

So it's just a bit more messy and on older systems we still get overmatching where the vid/pid/name are shared but overall it Should Just Work<sup>tm</sup> :)

data/huion-gc610-710.tablet
JoseExposito: I (finally) added your files for all the description files. Resolving this conversation ;)

data/huion-gc610-710.tablet
```diff
@@ -0,0 +1,26 @@
+# This tablet description file has been generated using an automated tool:
+# https://github.com/JoseExposito/uclogic-libwacom-tablet-descriptions
+# If you own this tablet, please improve it.
+
+[Device]
+Name=Huion GC610/710
```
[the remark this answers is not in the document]
JoseExposito: I forgot about this one. Should I add an "Unverified" at the beginning of the name?

JoseExposito:
> To illustrate my comment above: the only .tablet file that is overwritten by the autogenerated results in https://github.com/JoseExposito/uclogic-libwacom-tablet-descriptions/pull/1 is the HS611

Ops, I didn't automate the removal of devices already present in libwacom and instead I manually deleted them... And manual processes tend to have errors. That file shouldn't be there, I removed it in the latest revision.

This PR should contain only new tablets. If the tablet was already present in libwacom, we should use them.

JoseExposito:
Added an extra commit with 70 tablets (that should be all tablets) from GAOMON.

data/huion-gc610-710.tablet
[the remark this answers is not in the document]
whot: eh, thought about this and I don't think it's worth it. Any mistakes users will (hopefully) file a bug, otherwise there's no need to advertise it.

whot:
Alrighty, I've merged all the other PRs with the groundwork which means you should be able to rebase and get a) good enough testing and b) useful error messages for the SVG validity tester.

> Added an extra commit with 70 tablets (that should be all tablets) from GAOMON.

awesome, thanks. I'm not looking forward to adding all those to the NEWS file :)

JoseExposito:
Thanks a lot Peter, after rebasing on top of master CI is happy :tada:

PR #661, opened 2024-04-21
test: remove the requirement for a Layout file

whot:
Previously we required a layout file for any tablet with pad buttons. This somewhat prevents us from adding tablet files for unknown tablets where we can get most of the information from external sources, e.g. see #659. As a result we're missing support for these tablets altogether when we could have at least partial support.

Let's drop this requirement in our tests, users who want their tablet working properly can submit layout files.

cc @JoseExposito

### DIFF
--- a/test/test-tablet-svg-validity.c
+++ b/test/test-tablet-svg-validity.c
@@ -234,19 +234,6 @@ struct fixture {
 };
 
 static void
-test_filename(struct fixture *f, gconstpointer data)
-{
-    const WacomDevice *device = data;
-    const char *filename;
-
-    filename = libwacom_get_layout_filename(device);
-    if (libwacom_get_num_buttons(device) > 0) {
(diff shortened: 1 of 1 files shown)


JoseExposito (commented):
Mmmm what if I prefix the files added in https://github.com/linuxwacom/libwacom/pull/659 with `unverified-` and we only skip this test in that case?

Sorry for the CI failures, I'm not familiar with the tests yet.

whot:
> Mmmm what if I prefix the files added in https://github.com/linuxwacom/libwacom/pull/659 with unverified- and we only skip this test in that case?

Nah, too much effort I reckon. Might as well punt some of that to those that have the tablets :)

whot:
Going to merge this, #659 as a bigger and more beneficial long-term impact than disabling the tests.

PR #664, opened 2024-04-22
data: XP pen tablets have no Rings

whot:
They have a physical control that looks like a ring but it's not the Ring as our terminology has it. In the future we may add support for this as "Dial" but for now let's disable it.


JoseExposito (approved):
LGTM. These devices have a relative wheel, not an absolute one.

Issue #667, opened 2024-04-23
Janitor: .tablet files should have a NumRings entry


whot:
We currently have `Ring=true` and `Ring2=true` but for touchstrips we have `NumTouchstrips`. We should switch the `.tablet` files use `NumRings=2` instead for consistency.

PR #669, opened 2024-04-23
Use NumRings instead of separate Ring/Ring2

### LINKED ISSUE #668: Janitor: Add libwacom_get_num_rings()
And deprecate `libwacom_has_ring()` and `libwacom_has_ring2()` so we have a more consistent API.

whot:
This deprecates libwacom_has_ring() and libwacom_has_ring2() for the new libwacom_get_num_rings(). In our tablet files we also remove Ring=true/Ring2=true and replace that with NumRings=.

This makes the support line up better with strip support and removes some of the confusion in the tablet file where we have two independent Ring and Ring2 entries with different meanings.

Closes #668 #667


whot:
@JoseExposito sorry, this will probably conflict with #659

PR #673, opened 2024-04-25
Add support for Wacom Movink 

flying-elephant:
Those commits are ready for Wacom Movink announced yesterday.

### DIFF
--- a/data/libwacom.stylus
+++ b/data/libwacom.stylus
@@ -766,6 +766,6 @@ Name=UD Pen Eraser
 Group=udpen
 PairedStylusIds=0x10002;
 Buttons=1
-EraserType=Invert;
+EraserType=Invert
 Axes=Tilt;Pressure;Distance;
 Type=General
--- a/data/libwacom.stylus
+++ b/data/libwacom.stylus
@@ -756,7 +756,7 @@ Name=UD Pen
 Group=udpen
 PairedStylusIds=0x1000a;
 Buttons=2
-EraserType=Invert;
+EraserType=Invert
 Axes=Tilt;Pressure;Distance;
 Type=General
 
--- a/data/libwacom.stylus
+++ b/data/libwacom.stylus
@@ -749,3 +749,22 @@ Type=Puck
 HasLens=false
 HasWheel=true
 Buttons=5
+
+[0x10002]
+# Movink
+Name=UD Pen
+Group=udpen
+PairedStylusIds=0x1000a;
+Buttons=2
+EraserType=Invert;
+Axes=Tilt;Pressure;Distance;
+Type=General
+
+[0x1000a]
+# Movink
+Name=UD Pen Eraser
+Group=udpen
+PairedStylusIds=0x10002;
+EraserType=Invert;
+Axes=Tilt;Pressure;Distance;
+Type=General
--- a/data/libwacom.stylus
+++ b/data/libwacom.stylus
@@ -756,7 +756,6 @@ Name=UD Pen
 Group=udpen
 PairedStylusIds=0x1000a;
 Buttons=2
-EraserType=Invert
 Axes=Tilt;Pressure;Distance;
 Type=General
 
--- a/data/libwacom.stylus
+++ b/data/libwacom.stylus
@@ -765,6 +765,7 @@ Type=General
 Name=UD Pen Eraser
 Group=udpen
 PairedStylusIds=0x10002;
+Buttons=1
 EraserType=Invert;
 Axes=Tilt;Pressure;Distance;
 Type=General
--- a/data/movink.tablet
+++ b/data/movink.tablet
@@ -23,7 +23,7 @@
 [Device]
 Name=Wacom Movink
 ModelName=DTH-135
-DeviceMatch=usb:056a:03f0
+DeviceMatch=usb|056a|03f0
 Class=Movink
 Width=11
 Height=8
--- a/data/movink.tablet
+++ b/data/movink.tablet
@@ -27,7 +27,7 @@
 Name=Wacom Movink
 ModelName=DTH-135K0C
 DeviceMatch=usb|056a|03f0
-Class=Movink
+Class=Cintiq
 Width=11
 Height=8
 Styli=@udpen;
--- a/data/movink.tablet
+++ b/data/movink.tablet
@@ -19,6 +19,11 @@
 #    |                                          |
 #    *------------------------------------------*
 #
+# Note: This tablet has 2 physical buttons (one left, one right)
+# configured via a firmware OSD but only one button can send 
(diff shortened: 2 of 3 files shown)


data/movink.tablet
```diff
@@ -0,0 +1,42 @@
...
+#    |                                          |
+#    *------------------------------------------*
+[Device]
+Name=Wacom Movink
+ModelName=DTH-135
+DeviceMatch=usb:056a:03f0
+Class=Movink
+Width=11
+Height=8
+Styli=@udpen;
```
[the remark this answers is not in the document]
flying-elephant: >This tablet is not compatible with current AES styli or any other previous styli? 
No, not at all. Movink doesn't work with AES styli because the used technology is much different from it.
Although the technology in udpen is just like that of ProPen or our other branded pens, it cannot work with our previous products. Therefore, adding "udpen' into "Styli" line is needed for Movink.

data/movink.tablet
```diff
@@ -0,0 +1,42 @@
...
+
+[Features]
+Stylus=true
+Reversible=false
+Touch=true
+Buttons=3
+
+[Buttons]
+Left=A;C
+Right=B;C
```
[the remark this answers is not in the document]
flying-elephant: >Top=C, right?
I'll work on this, thanks!
There are two buttons on each sides, so I thought they should be the sides.
"Tpo=C" makes much sense.

Yes, one of the physical buttons can be assigned for an express key, but it also can be other functionalities than the express key; one of them is "Touch on/off".

data/libwacom.stylus
```diff
@@ -749,3 +749,22 @@ Type=Puck
...
+Type=General
+
+[0x1000a]
+# Movink
+Name=UD Pen Eraser
+Group=udpen
+PairedStylusIds=0x10002;
+EraserType=Invert;
+Axes=Tilt;Pressure;Distance;
+Type=General
```
whot, line 770: Missing a `Buttons` line. Looks like the old pens all had 1 button for the eraser but I  no longer remember why this was the case... I think this predates the distinction between tip and buttons so `Buttons=1` is necessary here.

data/movink.tablet
```diff
@@ -0,0 +1,42 @@
+# Wacom
+# Movink 13HD touch
+# DTH-135
+#
+# Note: touch data comes through the usb:056a:03f0 interface
```
whot, line 5: this is missing the sysinfo line (see the bot comment)

data/movink.tablet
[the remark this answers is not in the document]
whot: > can be assigned for an express key, 

so these are software/firmware controlled buttons? that's going to be ... interesting, libwacom doesn't have a good way to express these for now.

Might be best to add a comment in the file here and then figure out how to really handle this later when we have the consumers (e.g. mutter + gnome-control-center) actually wanting deal with these buttons.

whot (changes requested):
Thanks for the update. CI still fails so at least two changes are still needed - please make sure to run `meson test -C builddir` locally, that'll run (most of) the test suite locally so it'll speed up finding the various nitpicks my human eyes won't see :)

flying-elephant:
I modified the codes and updated them.
However, I got the attached log by running the suggestion, "meson test -C".
Do you know what is wrong exactly?
The log looks like the lines which describe about the eraser button, but I'm not sure what I should do.
"UD pen" could have an eraser button, so it has to stick with it somehow.
[testlog.txt](https://github.com/linuxwacom/libwacom/files/15175576/testlog.txt)

data/libwacom.stylus
```diff
@@ -749,3 +749,23 @@ Type=Puck
...
+
+[0x1000a]
+# Movink
+Name=UD Pen Eraser
+Group=udpen
+PairedStylusIds=0x10002;
+Buttons=1
+EraserType=Invert;
```
whot, line 769:
```suggestion
EraserType=Invert
```

data/libwacom.stylus
```diff
@@ -749,3 +749,23 @@ Type=Puck
...
 HasWheel=true
 Buttons=5
+
+[0x10002]
+# Movink
+Name=UD Pen
+Group=udpen
+PairedStylusIds=0x1000a;
+Buttons=2
+EraserType=Invert;
```
whot, line 759:
```suggestion
EraserType=Invert
```
This is a single value item so it probably fails to parse this one

whot (commented):
I think the eraser parsing fails, fixing that should fix the CI

data/movink.tablet
```diff
@@ -0,0 +1,45 @@
...
+#
+# sysinfo.4XIn8KkXzp
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/376#issue-2270545906
+[Device]
+Name=Wacom Movink
+ModelName=DTH-135K0C
+DeviceMatch=usb|056a|03f0
+Class=Movink
```
whot, line 30:
```suggestion
Class=Cintiq
```

`Class` is deprecated, no-one uses it and this is close enough to a cintiq so let's go with that.

data/libwacom.stylus
```diff
@@ -749,3 +749,23 @@ Type=Puck
...
 HasWheel=true
 Buttons=5
+
+[0x10002]
+# Movink
+Name=UD Pen
+Group=udpen
+PairedStylusIds=0x1000a;
+Buttons=2
+EraserType=Invert
```
whot, line 759:
```suggestion
```

EraserType needs to be set on the eraser, not on the pen

data/movink.tablet
```diff
@@ -0,0 +1,45 @@
...
+#    |                  TABLET                  |
+#    |                                          |
+#    |                                          |
+#    |                                          |
+#    |                                          |
+#    |                                          |
+#    *------------------------------------------*
+#
+# Note: touch data comes through the usb:056a:03f0 interface
+#
```
whot, line 23:
```suggestion
# Note: This tablet has 2 physical buttons (one left, one right)
# configured via a firmware OSD but only one button can send 
# events at any time. This is button C, shown centered here.
# We cannot tell which physical button triggered the event.
#
# Note: touch data comes through the usb:056a:03f0 interface
#
```

data/movink.tablet
```diff
@@ -0,0 +1,45 @@
...
+Layout=movink.svg
+
+[Features]
+Stylus=true
+Reversible=false
+Touch=true
+Buttons=3
+
+[Buttons]
+Left=A;C
```
whot, line 44:
```suggestion
# Button C can be physically left or right depending on the firmware
# but let's assign it to Left here, that's good enough.
Left=A;C
```

whot (approved):

whot:
All good now, let's get this merged! Thanks heaps for the effort

PR #674, opened 2024-04-28
Added more device names for kamvas 13 variants

LostNulide:
I am using the Kamvas 13 with the same model number, but it couldn't be detected by libwacom.
After some searching, I found out that the device name used in the descriptor doesn't match the one reported by my tablet.
I suspect there may be a different firmware installed.
Nevertheless, by adding the device names to the descriptor, everything is working.

Is it fine to use colon (:) as serperator?
The Pipe Symbol did not work on my distribution.
If not i can switch it to the other symbol.

### DIFF
--- a/data/huion-kamvas-13.tablet
+++ b/data/huion-kamvas-13.tablet
@@ -24,7 +24,7 @@
 [Device]
 Name=Huion Kamvas 13
 ModelName=GS1331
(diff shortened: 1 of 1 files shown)


whot:
The switch from `:` to `|` was a recent one (767ae7ecaaf47213bce967dfce115ed6a6f9118f) and hasn't been in a release yet. For this PR it needs to be the pipe symbol please, you can keep using a colon for your local overrides file.

LostNulide:
Got it. I switched to the pipe symbol.

whot:
merged, thanks

PR #675, opened 2024-04-29
meson.build: increase timeout of pytest test

alyssais:
Running pytest can take a while, especially if the system is already under load (for example, if it's building other packages at the same time as libwacom).  Recently, I saw it take 82 seconds.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -294,7 +294,8 @@ if get_option('tests').enabled()
 	     pytest,
 	     args: ['--verbose', '--log-level=DEBUG', meson.current_source_dir()],
 	     env: env,
-	     suite: ['all'])
+	     suite: ['all'],
+	     timeout: 300)
 endif
 
 # This is a non-optional test


whot:
> Recently, I saw it take 82 seconds.

This seems odd, my system here (not a particularly fast one...) finishes in under 4 seconds. I'd understand double that or triple that if the machine is under load but 82 seconds feels very wrong. 

There is one instance where it's asking for the password though (during test_hwdb_files I think) so maybe it's just hanging in that test?

alyssais:
It did eventually pass after 82 seconds — would that be expected if it was hanging?

whot:
tbh I don't know how long the timeout would be. If you run it as `pytest -v` what does the output look like? Do you see where it's hanging?

whot:
fwiw, e4ea346eb1fe51b401d068f8222d696e60d196fd may help too with the timeouts.

alyssais:
I'll try it next time I'm building libwacom — thanks!

whot:
Apologies, this is indeed an issue with the tests, I just didn't noticed because I either ran a subset of them or they all got skipped. Short answer: running systemd-hwdb and others on every test run is simply slow, so the whole `test_udev_rules.py` file takes forever.

whot:
Closing this in favour of #694 which is a better fix since the test suite no longer takes forever with that fix. We might still need a timeout (e.g. it can take 22s on my machine which is close to the 30s timeout) but we shouldn't need a huge increase.

PR #676, opened 2024-05-03
Add libwacom_new_from_builder()

### LINKED ISSUE #670: Debugging functions needed for creating tablets from matches
Our `libwacom_new*` functions don't allow for creating a libwacom device from a match string - this makes debugging harder especially where devices share vid/pids and the actual device isn't available locally.

Since the match string is an implementation detail this would have to be a private function or alternatively we could do some builder-like pattern to add vid/pid/bustype/name/uniq and pass that into a new public `libwacom_new` function.

See #666 where this would be handy to debug a device that doesn't exist locally.

whot:
This adds a new (and hopefully final) `libwacom_new_from...` call. This one takes a ~~match~~ `WacomBuilder` object which can now be constructed and filled by the user. In the future if we need to add further fields we can extend the `WacomBuilder` with a setter and have this function still work for us.

Notably, see https://github.com/linuxwacom/libwacom/pull/676#issuecomment-2092747045 the builder differentiates between the device name and the match name since those two may not be the same. The behaviour of the builder approach matches the existing approaches we have so the various `libwacom_new_from_` functions can be emulated with the new builder approach.

Prime motivation here is to be able to test better, I don't think any of our callers will switch since they all use `new_from_path`. In particular for the correct loading behaviour in #659.

Closes #670

### DIFF
--- a/libwacom/libwacom.c
+++ b/libwacom/libwacom.c
@@ -26,6 +26,7 @@
 
 #include "config.h"
 
+#include "libwacom.h"
 #include "libwacomint.h"
 #include <assert.h>
 #include <stdlib.h>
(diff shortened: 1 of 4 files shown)


whot:
One difficulty I just realised: `wacom_match_get_name()` returns the name string from the match but in this version I'm using `wacom_match_set_name()` as if it was the `WacomDevice->name`. That works for some tablets but won't work for all the Huion and derivatives which have different kernel names than we set in the `.tablet` files.

Might need to introduce a new struct other than `WacomMatch` that can take a `device_name` and a `match_name`.

whot:
Going to merge this, I think it's a useful (and extensible) API and it makes testing a lot easier.

PR #677, opened 2024-05-03
Add a python wrapper to test with pytest

whot:
Includes the equivalent to `test-load.c` and sits on top of #676 because it'll make testing that a bit simpler (which was the prime motivation for this function anyway).

Adding a python test suite is going to make testing things *a lot* easier, pytest is significantly more powerful than the glib test scaffolding and with the right wrappers adding things is a breeze.

The only files that matter here are `meson.build` and `test_libwacom.py`, the rest is from !676.

Main drawback: we can't test the new libwacom approach with asan/ubsan enabled, it crashes somewhere inside ctypes. Maybe that's fixable but right now pytest detects this (env var check) and skips those tests.


whot:
Merging this one too, with the svg test added this is getting quite useful and #678 shows that handling these tests in python is significantly more convenient than writing them in C. It's just our internal test suite anyway so we can change it at will (e.g. not fully happy with some of the enums in the python code but that can be fixed)

PR #678, opened 2024-05-03
Allow for a uniq-only match

whot:
First commit extracted from @JoseExposito's patchset in #659 but I think this needs a behaviour change to allow for uniq-only matches to take precedence over name matches, see https://github.com/linuxwacom/libwacom/pull/659#issuecomment-2091928456

Draft because it's untested and to test this we should add #677 and #676 first so we can test this properly.


whot:
With tests now, how awesome is that?! :)

JoseExposito:
Pretty awesome! I trust you on the automated testing side, but the uniq matching looks great.

PR #686, opened 2024-05-13
data: fix tablet file for the Huion Inspiroy 2S

whot:
PID is `0066`,~~ but filing as draft otherwise because I need to run this against a kernel with @JoseExposito's uniq patches.~~

TODO:
- [x] SVG update for rel dials
- [x] verify firmware version is correct - `E: HID_UNIQ=HUION_T21j_221221` with https://github.com/DIGImend/digimend-kernel-drivers/pull/690


JoseExposito (approved):
LGTM. Does it look like HUION is not reusing IDs with the latest devices? 🤔

whot:
:cautiously optimistic:

PR #687, opened 2024-05-13
Fix the pytest invocation

whot:
meson always sets `ASAN_OPTIONS` so the SVG tests didn't actually run in our CI. oops.

Filing as draft because pretty much all the SVGs added in #659 are missing some bits so they need to get fixed one-by-one.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -285,23 +285,25 @@ if get_option('tests').enabled()
 		message('valgrind not found, disabling valgrind test suite')
 	endif
 
-	env = environment()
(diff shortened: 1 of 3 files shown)


whot:
ftr, this change still skips the hwdb checks in the CI because of
```
SKIPPED [1092] ../test/test_udev_rules.py:117: Skipping test: [Errno 13] Permission denied: '/etc/udev/hwdb.d/99-libwacom-pytest.hwdb'
```

Need to fix that separately.

PR #690, opened 2024-05-14
Fix some issues with the recently added SVG files

whot:
-  Remove SVG files for tablets without rings/buttons/strips/dials. Those are just a single rect, we don't need those
- Remove a `,` separating the groups, this looks like a leftover from a runaway script
- Remove a `rect` that snuck its way into the Huion Mini Keydial

Related to #659, cc @JoseExposito


JoseExposito:
> Remove a , separating the groups, this looks like a leftover from a runaway script

Thanks! I removed it from the generator script in case we need to run it in the future.

PR #691, opened 2024-05-14
layouts: Fix buttons for Huion and Gaomon tablets

whot:
libwacom's SVG format requires that we have an element with
id="ButtonA", one with id="LabelA" and one with id="LeaderA".

The latter is the line pointing to the button. Since OSDs mostly fill in
the button rect when the button is pressed we need the label to be
outside the button and a leader pointing to it.

This patch was (supervised but) generated from a Python script:
all labels are moved to the right and a leader with the right id and
class was inserted. The script supports moving buttons (and adding
leaders) in all four directions, most tablets are ok now. Some tablets
have a less-than optimal layout (e.g. the Huion Mini Keypad) but that
can be fixed by someone with more time and motivation.

Script source:
https://gist.github.com/whot/8f3dbc09908862775b61f7d0a71fd5ed

Fixes: https://github.com/linuxwacom/libwacom/commit/c80bd7f7bfdfec63a70c0081c13f433854b103b9 ("Add auto generated GAOMON tablet description files")
Fixes: https://github.com/linuxwacom/libwacom/commit/ce093e730fb1901dad7b6bc259b0269bb4ca9a13 ("Add auto generated HUION tablet description files")

---

Related to #659, cc @JoseExposito 

Not all SVGs are great, arguably the original format with the button inside is better but unfortunately this is a change to how everything is set up. For now we need the leader lines (and probably someone to fix the various files that are going to be messy in the OSD).

### DIFF
--- a/data/layouts/gaomon-1060pro.svg
+++ b/data/layouts/gaomon-1060pro.svg
@@ -1,168 +1,66 @@
 <?xml version="1.0" standalone="no"?>
 <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
-<svg xmlns="http://www.w3.org/2000/svg"
-        version="1.1"
-        style="color:#000000;stroke:#7f7f7f;fill:none;stroke-width:.25;font-size:8"
-        id="gaomon-1060pro"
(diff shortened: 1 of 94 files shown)


JoseExposito (approved):
Looks great. The design is not perfect, but it'll incentive tablet owners to fix it.

PR #694, opened 2024-05-14
test: Replace the uinput-based hwdb test with a direct query

whot:
Previously we'd create a uinput device for each of our tablets, then
query udev if the property is set. This takes a long time, in parts
thanks to the 300ms sleep to wait for the uinput device to be ready.

With now over 500 tablet files and most of those having multiple matches
we're looking at at minimum 500*0.3 seconds for a test run which isn't
really useful for a test like this.

Instead - let's query systemd-hwdb directly: construct a query string
our udev rule and query systemd-hwdb for the properties, then ensure we
have the ones we expect.

Closes #675

cc @alyssais

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ permissions:
 
 env:
   CFLAGS: "-Werror -Wno-error=missing-field-initializers"
-  UBUNTU_PACKAGES: libgudev-1.0-dev libxml++2.6-dev valgrind tree python3-pip python3-setuptools libevdev-dev
+  UBUNTU_PACKAGES: libgudev-1.0-dev libxml++2.6-dev valgrind tree python3-pip python3-setuptools libevdev-dev udev
   PIP_PACKAGES: meson ninja libevdev pyudev pytest yq
 
 jobs:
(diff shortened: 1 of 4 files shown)


philiptaron:
@whot: Greetings! I hope you're well. Over in nixpkgs, we're still having flake reports on this test after this PR.

https://github.com/NixOS/nixpkgs/issues/328140#issuecomment-2260907269

CC: @eclairevoyant

whot:
@philiptaron: does increasing the timeout fix it?

philiptaron:
> @philiptaron: does increasing the timeout fix it?

I've never been able to consistently trigger it. Based on experience with valgrind in other projects, I've seen valgrind take much more time on loaded systems than unloaded ones, and every time I've gotten a build failure due to this test, the system has been under pretty heavy load. (For instance: compiling llvm, gcc, and other programs simultaneously.)

So, I don't know, sadly. When I do try to rebuild, it always works.

whot:
https://github.com/linuxwacom/libwacom/pull/757 may fix this, hopefully (if only by not running the pytests in valgrind anymore)

PR #699, opened 2024-05-16
Improve Huion Inspiroy 2 m support

mattie20:
- .tablet
    - Added button and dial map
    - Changed layout to be consistent with the design documents ( wacom.example )
    - Remove uniq from DeviceMatch field
- .svg
    - Change leaders to give more room. Prevent text and objects overlapping
    - Make the buttons more like the device/aesthetically pleasing
    - Organize and align text elements
    - Run through clean_svg.py

PR #700, opened 2024-05-16
data: update the Inspiroy 2L

tequeter:
Here is the updated `.tablet` file, but so far it doesn't pass the `libwacom-list-local-devices` test.

```
root@debian:~/libwacom# /usr/local/bin/libwacom-list-local-devices
/dev/input/event17 is a tablet but not supported by libwacom
/dev/input/event19 is a tablet but not supported by libwacom
Failed to find any devices known to libwacom.
```

It might be caused by the environment:

- The pipe-based DeviceMatch format isn't supported by the libwacom 2.6.0 of Debian stable.
- Somehow libwacom 2.10.0 of the Debian testing livecd I booted doesn't accept it either.
- So I compiled the master branch in that environment and installed it in the default path (mostly `/usr/local`).

This fixes the issue with the DeviceMatch format, but something is wrong with the resulting hwdb and/or support files.

I'm surprised that `/usr/local/lib/udev/rules.d/65-libwacom.rules` does a lookup on the name field while the generated `.hwdb` mostly has `input:usbid` information, but it's where I ran out of time.

[udevadm-test.txt](https://github.com/linuxwacom/libwacom/files/15341595/udevadm-test.txt)

I most likely won't be able to look into it again before next Wednesday, so here is this intermediate update so that you're not left hanging.


whot:
thanks!

fwiw, you can steal the SVG from #699 (cc @mattie20) - we can use the same file for both since they're almost the same (let's copy it though please instead of both `.tablet` files linking to the same svg). It'd be great if you could add this to this PR.

> The pipe-based DeviceMatch format isn't supported by the libwacom 2.6.0 of Debian stable.

yeah, if you change it locally to `DeviceMatch=usb:256c:0068::` it should work on your box, we don't have a libwacom release with the pipe syntax (or fw matching) yet.

This device seems to have its own pid too so you could remove the uniq match since we don't need it (let's hope that remains true for the future) and it'll make the match work immediately rather than on 6.10+ kernels only.

> I'm surprised that /usr/local/lib/udev/rules.d/65-libwacom.rules does a lookup on the name field while the generated .hwdb mostly has input:usbid information, but it's where I ran out of time.

that's a bit of a misunderstanding. the udev rule adds a *prefix* to the modalias. that prefix contains the name of the device so the full lookup string for any device would then be:
```
libwacom:name:Your Device Name:input:b0003v1234p56AB....
```
(see `/sys/class/input/event6/device/modalias` for whichever event node you want to check).

All our hwdb matches only match on name and vid/pid and we don't care about the gobbledygook after, but in theory we could make a hwdb match on "does device that have `KEY_NUMLOCK`".

tequeter:
Indeed, it's easier like this! Now this works:

```
$ libwacom-list-local-devices                                                                     
devices:
- name: 'Huion Inspiroy 2 L'
  bus: 'usb'
  vid: '0x256c'
  pid: '0x0068'
  nodes: 
  - /dev/input/event12
  - /dev/input/event11
  - /dev/input/event10
  - /dev/input/event9
  - /dev/input/event5
  - /dev/input/event4
```

Please note that the above was obtained with a different `.tablet` than this PR's, one that was modified to fit my libwacom-2.6.0. I suggest that you review the differences in https://github.com/tequeter/libwacom/commit/a8eb299c98444cb74c60803578c35848098af0ab to make sure that the one in the PR will work, given a recent enough libwacom.

I also replaced the layout file using @mattie20 's nicer SVG as requested.

Is there anything else to test?

data/huion-inspiroy-2-l---h1061p.tablet
```diff
@@ -1,17 +1,43 @@
...
 [Device]
-Name=Inspiroy 2 L - H1061P
-ModelName=Inspiroy 2 L - H1061P
+Name=Huion Inspiroy 2 L
+ModelName=H1061P
 Class=Bamboo
-DeviceMatch=usb|256c|0068||HUION_T21m
-Width=9  # autogenerated value, please fix
-Height=6  # autogenerated value, please fix
+DeviceMatch=usb:256c:0068;
```
whot, line 38:
```suggestion
DeviceMatch=usb|256c|0068;
```

data/huion-inspiroy-2-l---h1061p.tablet
```diff
@@ -24,7 +50,8 @@ Reversible=true
 Stylus=true
 Touch=false
 TouchSwitch=false
+Buttons=11
 
 [Buttons]
 Left=A;B;C;D;E;F;G;H;I;J;K
-EvdevCodes=BTN_0;BTN_1;BTN_2;BTN_3;BTN_4;BTN_5;BTN_6;BTN_7;BTN_8;BTN_9;BTN_SOUTH
+EvdevCodes=0x100;0x101;0x102;0x103;0x104;0x105;0x106;0x107;0x108;0x109;0x130
```
whot, line 57:
```suggestion
EvdevCodes=BTN_0;BTN_1;BTN_2;BTN_3;BTN_4;BTN_5;BTN_6;BTN_7;BTN_8;BTN_9;BTN_SOUTH
```

whot (approved):
Undoing the 2.6 changes but otherwise this LGTM, thanks.

whot:
And we're good to go, thanks

PR #703, opened 2024-05-20
Huion RTM-500 Support

mattie20:
- .tablet
    - Format file according to example.wacom
    - Add sysinfo reference
    - Add Button map
    - Add proper PID to DeviceMatch field
- .svg
    - Add touchstrip
    - Align and space leaders, objects, and text
    - Run through clean_svg.py


data/huion-rtm-500.tablet
```diff
@@ -1,29 +1,59 @@
...
 [Device]
 Name=Huion RTM-500
-ModelName=RTM-500
+ModelName=H952
 Class=Bamboo
-# Product ID is unknown, please fix
-DeviceMatch=usb|256c|006e||HUION_T19h;usb|256c|006d||HUION_T19h;usb|256c|006f||HUION_T19h;usb|256c|0064||HUION_T19h;
-Width=9  # autogenerated value, please fix
-Height=6  # autogenerated value, please fix
+DeviceMatch=usb|256c|006d;
```
whot, line 42: You'll need to keep the firmware match here, we already have devices with 006d, it's one of the re-used PIDs. see e.g. the H1060p and the Gaomon S620.

So this line needs to be:
```suggestion
DeviceMatch=usb|256c|006d||HUION_T19h;
```

If it has a unique-enough name we can add that though.

mattie20: The name is pretty much just the tablet name. Strangely when I tried to search through their linux driver, I couldn't find many references to PID's in files or function calls. They do keep a json of their product mappings, but only in reference to those "HUION_T19h" type aliases.

whot: Yeah, huion (and thus gaomon/uclogic) have been re-using the PIDs for a while now which has been an issue for quite a while for libwacom. In #659 we added all the ones extracted from the driver (and the kernel's hid-uclogic.ko now provides the firmware version as the HID device's uniq so libwacom can access it).

At least once kernel 6.10 comes out this should be mostly fixed and we don't need the name matches anymore because we can match against the uniq now.

On the positive side, at least the recent Inspiroy 2 series now uses individual PIDs so maybe this is a thing of the past now.

whot:
merged, thank you

PR #704, opened 2024-05-24
Add support for Huion WH1409

crimsonfall:
Hello!

These are the two files (.tablet and .svg) to add support to the [Huion WH1409](https://www.huion.com/pen_tablet/Inspiroy/WH1409(8192).html). You can find the link to the sysinfo tarball within the .tablet file, or over [here](https://github.com/linuxwacom/wacom-hid-descriptors/issues/385).

Just for clarity, this is **not** the V2 version of the tablet.

I also need to state that while `libwacom-list-local-devices` shows the device correctly, KDE Tablet Settings, at least under X11, does not currently pick it up. Actually it picks the tablet up as a game controller, which is odd. I don't know if that bit of information is useful or not. Otherwise the tablet, including pen pressure, works just fine.

If you have any concerns about my pull request or anything regarding the last point, please let me know!

Thanks!

### DIFF
--- a/data/huion-wh1409.tablet
+++ b/data/huion-wh1409.tablet
@@ -21,7 +21,7 @@
 [Device]
 Name=TABLET Pen Tablet
 ModelName=
-DeviceMatch=usb|256c|006e|TABLET Pen Tablet Pad;usb|256c|006e|TABLET Pen Tablet Pen;usb|256c|006e|TABLET Pen Tablet Touch Strip;usb|256c|006e|TABLET Pen Tablet Dial
+DeviceMatch=usb|256c|006e||HUION_T153;usb|256c|006e|TABLET Pen Tablet Pad;usb|256c|006e|TABLET Pen Tablet Pen;usb|256c|006e|TABLET Pen Tablet Touch Strip;usb|256c|006e|TABLET Pen Tablet Dial
 Class=Bamboo
 Width=14
 Height=9
--- a/data/huion-wh1409.tablet
+++ b/data/huion-wh1409.tablet
@@ -1,0 +1,40 @@
+# HUION
+# WH1409
+#
+# sysinfo.DFvHUfcRjW.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/385
+#
+# Button Map:
+# (A=1, B=2, C=3, ...)
+#
+#         *----------------------------*
+#    A B  |                            |
+#    c D  |                            |
+#         |                            |
+#    E F  |           TABLET           |
+#    G H  |                            |
+#         |                            |
+#    I J  |                            |
+#    K L  |                            |
+#         *----------------------------*
+
+[Device]
+Name=TABLET Pen Tablet
+ModelName=
+DeviceMatch=usb|256c|006e|TABLET Pen Tablet Pad;usb|256c|006e|TABLET Pen Tablet Pen;usb|256c|006e|TABLET Pen Tablet Touch Strip;usb|256c|006e|TABLET Pen Tablet Dial
+Class=Bamboo
+Width=14
+Height=9
+IntegratedIn=
+Layout=huion-wh1409.svg
+Styli=@generic-no-eraser;
+
+[Features]
+Stylus=true
+Reversible=true
+Touch=false
+NumStrips=0
+
+[Buttons]
+Left=A;B;C;D;E;F;G;H;I;J;K;L
+EvdevCodes=BTN_0;BTN_1;BTN_2;BTN_3;BTN_4;BTN_5;BTN_6;BTN_7;BTN_8;BTN_9;BTN_SOUTH;BTN_EAST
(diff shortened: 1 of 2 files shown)


data/huion-wh1409.tablet
```diff
@@ -0,0 +1,40 @@
...
+#         |                            |
+#    E F  |           TABLET           |
+#    G H  |                            |
+#         |                            |
+#    I J  |                            |
+#    K L  |                            |
+#         *----------------------------*
+
+[Device]
+Name=TABLET Pen Tablet
```
whot, line 22:
```suggestion
Name=Huion WH1409
```
fwiw, the purpose of the `Name=` field is to give us a nice human-readable name that's better than the garbage the firmware gives us.

data/huion-wh1409.tablet
```diff
@@ -0,0 +1,40 @@
...
+ModelName=
+DeviceMatch=usb|256c|006e|TABLET Pen Tablet Pad;usb|256c|006e|TABLET Pen Tablet Pen;usb|256c|006e|TABLET Pen Tablet Touch Strip;usb|256c|006e|TABLET Pen Tablet Dial
```
whot, line 24: tbh, I'm a bit hesitant to merge this because it's probably going to conflict with other devices that re-use the same PID. As it happens this is the first device with his particular name match but still...

I'm guessing you're using the digimend drivers for this tablet? What's the value of HID_UNIQ for this device, that should contain the firmware string (`udevadm info --export-db` then search for the tablet).

whot:
>  Actually it picks the tablet up as a game controller, which is odd. I

Might be that `ID_INPUT_JOYSTICK` is set but that should be unset once libwacom has been installed and you've re-plugged the device. `libinput record` will show the udev properties, what's the output of that?

whot (commented):
Thanks, this mostly looks good but we will need the fw ID. And I'll have to make up my mind about merging name-only matches...

crimsonfall:
Hey!

So I ran the above udevadm command, however I got the following as a result for my tablet.

`E: HID_NAME=TABLET Pen Tablet`
`E: HID_PHYS=usb-0000:02:00.0-6.3/input0`
`E: HID_UNIQ=`

The HID_UNIQ was completely empty. Even when I ran the command under sudo, it still doesn't show up. In fact , the only device I have that would show the HID_UNIQ for was my mouse.

I apologize if this doesn't help with things.

For the record, I didn't explicitly install the Digimend drivers; I'd assume it came with my kernel on Fedora. It appeared to work because for a while, I was using a Xorg conf file to get the tablet to be detected by xsetwacom and was even able to configure the buttons. In addition, KDE Tablet Settings did detect the tablet, but I couldn't configure the buttons through the GUI since they wouldn't apply so I had to use the xsetwacom tool to do so.  Again, I don't know if that bit of info is useful or not.

whot:
If the tablet is a bit older already then it's probable that the digimend drivers got upstreamed and are in fedora. It'd still be nice if you could try it with current digimend master because that extracts the firmware version. You can build it, `rmmod hid-uclogic; insmod hid-uclogic.ko`, plug your device in and the HID_UNIQ should be set. No need to install and rebooting gets you back to normal.

Having said that: [this tool](https://github.com/whot/huion-switcher) should print it too though I'm not sure if it'll work if the driver has initialized the tablet already.

data/huion-wh1409.tablet
whot, line 24:
```suggestion
DeviceMatch=usb|256c|006e||HUION_T153;usb|256c|006e|TABLET Pen Tablet Pad;usb|256c|006e|TABLET Pen Tablet Pen;usb|256c|006e|TABLET Pen Tablet Touch Strip;usb|256c|006e|TABLET Pen Tablet Dial
```

whot:
Thanks, I added this as a match for the device. Since the uniq matches won't take effect until kernel 6.10 let's merge the name matches for this one too, you're lucky that you're the first with that particular name :)

crimsonfall:
Thanks! :)

> Might be that ID_INPUT_JOYSTICK is set but that should be unset once libwacom has been installed and you've re-plugged the device. libinput record will show the udev properties, what's the output of that?

I'm totally sorry for not answering this question you had earlier. So when running `libinput record` to the tablet, I get this under udev:

`udev:`
    `properties:`
   `- ID_INPUT=1`
   `- ID_INPUT_TABLET=1`
   `- ID_INPUT_TABLET_PAD=1`
   ` - LIBINPUT_DEVICE_GROUP=3/256c/6e:usb-0000:02:00.0-6`

So from the looks of it, libwacom is detecting it as a tablet, but for some reason when I go into KDE Tablet Settings, even after putting the data files into the libwacom folder, updating the libwacom database, and replugging my device, KDE Tablet Settings still doesn't see the tablet. At this point, I feel like this is a KDE specific issue.

If you still want to discuss this issue, we should probably do so outside of this PR discussion. Otherwise, I'll probably just file a bug with the KDE team and talk to them about it. Thanks for the merge though!

PR #708, opened 2024-05-28
Fetch the name and uniq from udev properties, not sysfs

whot:
Both are set by udev based on the sysattr entries anyway and using the properties instead of the sysfs attr means we can override both via udev rules should that become necessary.

cc @JoseExposito


JoseExposito (approved):
LGTM, better to have it in case we need it in the future :)

PR #710, opened 2024-05-29
Don't mark Keyboard and Mouse devices as ID_INPUT_TABLET or ID_INPUT_TABLET_PAD

whot:
This should cut down on false positives or at least the ones caused by us.

Related https://github.com/DIGImend/digimend-kernel-drivers/pull/695, cc @JoseExposito


JoseExposito (approved):
Thanks! I'm incredibly busy this week, but I'll try to test it with my device (and provide a recording) once I have some time.

PR #726, opened 2024-06-18
data: fix the Huion Kamvas 12 file

whot:


whot:
ftr, device works out of the box with Fedora 40. the size is off by a few cm each direction which makes the "keep aspect" ratio feel wrong but otherwise it's seems to work just fine.

PR #741, opened 2024-07-17
Id and group modifications

flying-elephant:
We found that there was the unnecessary ID remained on libwacom.stylus
and Movink supports more groups for libwacom to work properly.

### DIFF
--- a/data/libwacom.stylus
+++ b/data/libwacom.stylus
@@ -303,14 +303,6 @@ Buttons=3
 Axes=Tilt;Pressure;Distance;
 Type=General
 
-[0x40200]
-# Cintiq Pro 2022
-Name=Pro Pen 3
-Group=cintiqpro2022
-Buttons=3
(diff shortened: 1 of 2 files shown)


whot:
LGTM, thanks. did you mean to add a double-signed off by here? :)

flying-elephant:
The double Signed-offs are just in case.
Like LKML, an e-mail which is not matched to the registered e-mail in a system, I thought it might be sage to
provide two e-mails.
Does it lead any problem?

whot:
nah, no problem. here on github (same for gitlab) we don't send emails so the sender email doesn't matter and you can adjust the git committer email to match whatever signed-off-by tag you prefer. IOW there's no need for a double sign off.

flying-elephant:
I got it. I'll add an e-mail only matching my git committer e-mail. Thanks :)

PR #743, opened 2024-07-22
Update huion-kamvas-24-gs2401.tablet

InfiniteNightDreamer:
This vid:pid is also being used by  Huion Kamvas 13
data/huion-kamvas-pro-13.tablet


data/huion-kamvas-24-gs2401.tablet
```diff
@@ -1,18 +1,17 @@
...
 
 [Device]
 Name=KAMVAS 24 GS2401
 ModelName=KAMVAS 24 GS2401
 Class=Cintiq
-# Product ID is unknown, please fix
-DeviceMatch=usb|256c|006e||HUION_M206;usb|256c|006d||HUION_M206;usb|256c|006f||HUION_M206;usb|256c|0064||HUION_M206;
-Width=9  # autogenerated value, please fix
-Height=6  # autogenerated value, please fix
+DeviceMatch=usb|256c|006d;
```
whot, line 12: This we can't do - you're now matching on every single device with the 006d PID and potentially (well, definitely) breaking all other devices with the same PID. This needs to have the firmware specifier in it so it only matches on the kamvas 24.
```
DeviceMatch=usb|256c|006d||HUION_M206
```
alternatively/in addition if the device has a unique device name we can use that here too, see e.g. `data/huion-h640p.tablet`

whot (changes requested):

data/huion-kamvas-24-gs2401.tablet
InfiniteNightDreamer: So shoud it be like this ?
`DeviceMatch=usb|256c|006d|HUION Huion Tablet_GS2401 Tablet Monitor Pen;` ?
I assume we don't need Pad because there are no buttons on Kamvas 24 display tablet.
Kamvas 24 has built in monitor, looking at huion-kamvas-13.tablet file should this be also included ? 
`usb|256c|006d|Tablet Monitor;`

whot:
```
 DeviceMatch=usb|256c|0006d||HUION_M206;usb|256c|006d|HUION Huion Tablet_GS2401 Tablet Monitor Pen;
```
This way we have both the firmware match and the name match (the latter may prove to be less reliable although this one at least has the model number in the name).

I doubt the `tablet monitor` device doesn't need an entry - maybe the `Tablet Monitor Pen` if it sends events? Looking at the kamvas 13 file I'm not sure this is really needed tbt.

data/huion-kamvas-24-gs2401.tablet
```diff
@@ -6,10 +6,10 @@
 # https://github.com/linuxwacom/wacom-hid-descriptors/issues/398
 
 [Device]
-Name=KAMVAS 24 GS2401
-ModelName=KAMVAS 24 GS2401
+Name=KAMVAS 24
+ModelName=GS2401
 Class=Cintiq
-DeviceMatch=usb|256c|006d;
+DeviceMatch=usb|256c|006d|Huion Tablet_GS2401 Pen;usb|256c|006d|Huion Tablet_GS2401 Stylus;usb|256c|006d|Huion Tablet_GS2401 Pad;usb|256c|006d|Huion Tablet_GS2401;
```
whot, line 12:
```suggestion
DeviceMatch=usb|256c|006d||HUION_M206;usb|256c|006d|Huion Tablet_GS2401 Pen;usb|256c|006d|Huion Tablet_GS2401 Stylus;usb|256c|006d|Huion Tablet_GS2401 Pad;usb|256c|006d|Huion Tablet_GS2401;
```

whot:
merged, thanks

PR #752, opened 2024-08-09
data: Add stylus definition for Pro Pen 3E

jigpu:
The Pro Pen 3E is a version of the Pro Pen 3 which includes an eraser and is compatible with the same devices. This patch also adds the propen3 group to Movink, which should have had it.


whot (approved):

whot:
merged, thanks.

PR #753, opened 2024-08-12
data: Update comment header for 1st- and 2nd-gen Intuos Pro

jigpu:
Updates the comment headers for these devices to use the marketing name (already properly set in the "Name" field) and move the generation information to its own separate line).


whot:
I can see your marketing department is still strong with their product naming foo :) Merged, thanks.

PR #757, opened 2024-08-15
meson.build: bump the timeout for pytest to 60s and skip valgrinding python tests

whot:
This may or may not fix the issues with timeouts on the hwdb query tests, see #694.

On top of that skip all pytest for valgrind because there's no point.


philiptaron (approved):
Yeah, that looks right.

whot:
thanks, merged. let me know how you go with those in now

philiptaron:
We'll pull it with the next release tag. Thank you!

PR #770, opened 2024-09-29
data: add ELAN 9008 and 9009 (zenbook duo UX8406MA 1200p)

barskykd:
Adding .tablet files for both displays for zenbook duo UX8406MA 1200p-version


whot:
thanks!

PR #773, opened 2024-10-07
Minor adjustments for Lenovo Yoga 9 14IAP7

RayJW:
This is a follow-up to my previous PR, which just removes some old properties that still are present in the [wiki](https://github.com/linuxwacom/libwacom/wiki/Tablet-Definition-Files) but not needed any more according to the example file. I also renamed the file to fit with other similar devices and removed the Pen suffix for the Name property since both devices are handled with the same tablet definition.
Last but not least I added the Styli property, not sure if that changes anything, but it should be correct according to all the other Lenovo tablet definitions.

### DIFF
--- a/data/wacom-isdv4-52c2.tablet
+++ b/data/wacom-isdv4-52c2.tablet
@@ -4,18 +4,15 @@
 # https://github.com/linuxwacom/wacom-hid-descriptors/issues/227
 
 [Device]
-Name=Wacom HID 52C2 Pen
+Name=Wacom HID 52C2
 ModelName=WACF2200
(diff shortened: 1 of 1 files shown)


whot:
Please change the commit message to something like "data: fix outdated properties for Lenovo Yoga 9 14IAP7" - unfortunately I can't do this for a single commit MR. Thanks

whot:
oh, and I've fixed the wiki, at least that page, thanks

whot (approved):

whot:
merged, thanks

PR #782, opened 2024-10-11
Fix error in clean_svg.py and allow it to run on `.tablet` files

whot:
Fixes the crasher reported by @Deevad in https://github.com/linuxwacom/libwacom/pull/777#issuecomment-2405806463


Deevad:
Thank you for the quick fix! :+1:

PR #787, opened 2024-10-15
Point people to the better fleshed out wiki page for adding new devices

redstrate:
These instructions are old, and then I followed them because I didn't know there was a wiki page. Let's link that in the README so other people don't make my mistake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,18 +29,7 @@ device in the kernel (see `/proc/bus/input/devices`) and in the X session (see
 Use the `libwacom-list-devices` tool to list all known devices and verify
 the tablet is not in that list.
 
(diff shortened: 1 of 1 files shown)


redstrate:
Although now that I'm looking, it seems even those are outdated - it recommends adding a descriptor but that repository seems unmaintained at the moment?

whot (approved):

whot:
it's not unmaintained as such (well, it basically is but it doesn't matter), it's just a repository to keep things there. We should eventually merge all the open bug reports but meanwhile simply having those tarballs somewhere we can get to them later is enough.

Pointing to the wiki sounds like a plan, at least that one is easier to fix, thanks!

PR #788, opened 2024-10-15
data: add the XP-Pen ACK05 Remote, take 2

bentiss:
Taking over #777, because I have the HW locally and could do some more testing.

This has a couple of `clean_svg.py` fixes, and the SVG from @Deevad (fixed by me).

The dial still doesn't work properly in the overlay, so I wonder if the problem is not in gnome-settings-daemon. KDE only shows the buttons, not the dial.

I also tried adding the "modes", but it seems that they are not recognized by default by g-s-d (is this because there is no LEDs on the device and then non sysfs for them?)

Also, still no sysinfo, this is pretty much pre-development at thios stage, and we need to figure out if we are happy with the button mapping first.

### DIFF
--- a/data/xp-pen-ack05-remote.tablet
+++ b/data/xp-pen-ack05-remote.tablet
@@ -1,0 +1,41 @@
+# XP-Pen
+# ACK05 Remote
+#
+# Reports as Hanvon Ugee Shortcut Remote, for recordings
+# etc. see https://gitlab.freedesktop.org/libevdev/udev-hid-bpf/-/issues/32
+# and
+# sysinfo.PUwwSY53x2.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/415
+#
+#  ┌────────┐────────────────────────┐
+# /  ┌────┐  \   ┌───┐┌───┐┌───┐┌───┐│
+# │  │ A  │  │   │ B ││ C ││ D ││   ││
+# │  └────┘  │   └───┘└───┘└───┘│ H ││
+# \_________/    ┌───┐┌───┐┌───┐│   ││
+#  │             │ E ││ F ││ G ││   ││
+#  │             └───┘└───┘└───┘└───┘│
+#  │             ┌───┐┌────────┐┌───┐│
+#  │             │ I ││    J   ││ K ││
+#  │             └───┘└────────┘└───┘│
+#  └─────────────────────────────────┘
+
+[Device]
+Name=XP-Pen ACK05 Remote
+ModelName=
+DeviceMatch=usb|28bd|0202
+Layout=xp-pen-ack05-remote.svg
+Class=Remote
+
+[Features]
+Stylus=false
+NumDials=1
+NumRings=0
+NumStrips=0
+
+[Buttons]
+Left=A;B;C;D;E;F;G;H;I;J;K;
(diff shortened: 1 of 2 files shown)


tools/clean_svg.py
```diff
@@ -84,14 +84,21 @@ def round_if_number(value):
...
+    cleanup_elems = [
+        elem
+        for elem in root
+        if not elem.tag.startswith(BRACKETS_NAMESPACE)
+        or elem.tag == BRACKETS_NAMESPACE + "metadata"
+    ]
+    for elem in cleanup_elems:
+        root.remove(elem)
+
     for elem in root:
```
whot, line 100: heh, I think you could've gotten the same effect with:
```
for elem in list(root):
```
which *should* create a list first and then operate on the list instead.

data/xp-pen-ack05-remote.tablet
```diff
@@ -0,0 +1,39 @@
...
+DeviceMatch=usb|28bd|0202
+Layout=xp-pen-ack05-remote.svg
+Class=Remote
+
+[Features]
+Stylus=false
+NumDials=1
+NumRings=0
+NumStrips=0
+StatusLEDs=Ring
```
whot, line 31: If the LED doesn't light up on the device, we don't really have a status led, isn't it? Or am I misunderstanding something here?

whot:
> so I wonder if the problem is not in gnome-settings-daemon.

ftr, g-s-d is mostly dead and doesn't do much anymore, virtually all the code is now in gnome-control-center (read-only stuff) and mutter.

but yes, the dials won't work in GNOME yet because the wayland protocol is not upstream which means mutter etc. cannot use it.

tools/clean_svg.py
bentiss: indeed, it works

redstrate:
> The dial still doesn't work properly in the overlay, so I wonder if the problem is not in gnome-settings-daemon. KDE only shows the buttons, not the dial.

Work in progress, I'm eventually going to show the dial in the UI :)

bentiss:
> Work in progress, I'm eventually going to show the dial in the UI :)

Great news! Thanks a lot for working on this.

Currently the plan is to not merge things here and in udev-hid-bpf as long as things are not ready in KDE/Gnome. Simply because otherwise we are just removing the somehow useful Ctrl+/Ctrl- default output of the Diai and will break things for users.

data/xp-pen-ack05-remote.tablet
```diff
@@ -0,0 +1,41 @@
...
+# \_________/    ┌───┐┌───┐┌───┐│   ││
+#  │             │ E ││ F ││ G ││   ││
+#  │             └───┘└───┘└───┘└───┘│
+#  │             ┌───┐┌────────┐┌───┐│
+#  │             │ I ││    J   ││ K ││
+#  │             └───┘└────────┘└───┘│
+#  └─────────────────────────────────┘
+
+[Device]
+Name=Hanvon Ugee Shortcut Remote
```
whot, line 23:
```suggestion
Name=XP-Pen ACK05 Remote
```
The name is the one to display to the user, that the device reports itself as something different on the kernel doesn't matter, better to not confuse the users here.

If there is an actual ugee remote then we can change this.

whot (approved):
@bentisss: can you rebase this one please. I don't see why we cannot already merge this one - any `.tablet` file for this device will look the same, regardless of whether BPF is involved of not.

bentiss:
OK, rebased, fixed the tablet, force pushed and removing draft status

whot:
merged, thanks

PR #792, opened 2024-10-16
Generate the class given in class if not explicitly given

redstrate:
This is still a part of the API, and thus have required even new .tablet files to fill this out even though it's superseded by `IntegratedIn`. So let's generate them programmatically.

This turns devices that are `IntegratedIn=Display` to the `Cintiq` class, `IntegratedIn=Display;System` into the `ISDV4` class, and everything else into the `Bamboo` class.

(This definitely needs more test cases)

### DIFF
--- a/libwacom/libwacom.c
+++ b/libwacom/libwacom.c
@@ -1362,7 +1362,17 @@ libwacom_get_height(const WacomDevice *device)
 LIBWACOM_EXPORT WacomClass
 libwacom_get_class(const WacomDevice *device)
 {
-	return device->cls;
+	if (device->cls != WCLASS_UNKNOWN) {
+		return device->cls;
+	}
+	if (device->integration_flags & WACOM_DEVICE_INTEGRATED_DISPLAY &&
+		device->integration_flags & WACOM_DEVICE_INTEGRATED_SYSTEM) {
+		return WCLASS_ISDV4;
+	}
+	if (device->integration_flags & WACOM_DEVICE_INTEGRATED_DISPLAY) {
+		return WCLASS_CINTIQ;
+	}
+	return WCLASS_BAMBOO;
 }
 
 LIBWACOM_EXPORT int
--- a/data/wacom-cintiq-24hd-touch.tablet
+++ b/data/wacom-cintiq-24hd-touch.tablet
@@ -44,7 +44,6 @@ Name=Wacom Cintiq 24HD touch
 ModelName=DTH-2400
 DeviceMatch=usb|056a|00f8
 PairedID=usb|056a|00f6
-Class=Cintiq
 Width=21
 Height=13
 Layout=wacom-cintiq-24hd.svg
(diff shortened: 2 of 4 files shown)


libwacom/libwacom.c
```diff
@@ -1362,7 +1362,17 @@ libwacom_get_height(const WacomDevice *device)
...
+	if (device->cls != WCLASS_UNKNOWN) {
+		return device->cls;
+	}
+	if (device->integration_flags & WACOM_DEVICE_INTEGRATED_DISPLAY &&
+		device->integration_flags & WACOM_DEVICE_INTEGRATED_SYSTEM) {
+		return WCLASS_ISDV4;
+	}
+	if (device->integration_flags & WACOM_DEVICE_INTEGRATED_DISPLAY) {
+		return WCLASS_CINTIQ;
+	}
```
whot, line 1374:
```suggestion
	unsigned int flags = device->integration_flags & (WACOM_DEVICE_INTEGRATED_DISPLAY|WACOM_DEVICE_INTEGRATED_SYSTEM);
	
	if (flags == (WACOM_DEVICE_INTEGRATED_DISPLAY|WACOM_DEVICE_INTEGRATED_SYSTEM)) {
		return WCLASS_ISDV4;
	}
	if (flags == WACOM_DEVICE_INTEGRATED_DISPLAY) {
		return WCLASS_CINTIQ;
	}
```
is i think a bit more readable. 

(oh, and I always like to have `()` around any bitwise operations, it makes the code a bit more readable)

libwacom/libwacom.c
```diff
@@ -1362,7 +1362,17 @@ libwacom_get_height(const WacomDevice *device)
...
+	return WCLASS_BAMBOO;
```
whot, line 1375: I think the `Remote` class is missing here, that will need handling, could go together with #779.

Then again there's so few of them that we can mark those in the `.tablet` file...

whot (changes requested):
One minor style nitpick but this LGTM.

For the test cases - remove it from the `.tablet` files for the few other devices we hand-check in `test-load` and I think we'll be good there?

libwacom/libwacom.c
whot: Oh, come to think of it: this is better as `WCLASS_INTUOS5` - that's the most recent class and most devices are more like that than whatever the latest bamboo was like.

whot:
@redstrate ok if I take this one and push it over the line?

redstrate:
Yes go ahead, I don’t have access to a computer right now and would appreciate it :)On Oct 27, 2024, at 23:09, Peter Hutterer ***@***.***> wrote:﻿
@redstrate ok if I take this one and push it over the line?

libwacom/libwacom.c
whot: For the archives: `WCLASS_BAMBOO` is better here because bamboos arent' expected to have stylus ids/serials.

whot:
Thanks, closing in favour of #814 now

PR #793, opened 2024-10-16
Add support for Lenovo Precision Pen 2 (2023) Stylus

### LINKED ISSUE #774: Add Lenovo Precision Pen 2
Something changed in GNOME 47 and the non-eraser button on my stylus has become basically useless as it now simply acts as a left-click instead of correctly mapping to secondary actions in apps like Xournal++ or Rnote. I saw [this,](https://who-t.blogspot.com/2024/06/gnome-tablet-support-papercut-fixes.html) and I think one of the defaults that changed may have affected that behaviour.

I know that this is only vaguely related to libwacom, so this is just background info. Now the real question. I saw that there is now support for non-wacom styli. How can I help to get my stylus supported, so I can properly adjust the button mappings. At the moment, `sudo libwacom-show-stylus` simply prints the serial number and `Unknown stylus id 0x866b. New entry needed`.

I searched the wiki and all the other repos, but there doesn't seem to be any info about this for the time being, and the error only reads `Suggested line for .tablet file: Styli=` which doesn't help.

- **Device name**: Lenovo Precision Pen 2 (2023)
- **Device model identifier**: ZG38C04471 ?
- **libwacom version**: 2.13.0-1

- [X] I understand that libwacom does affect whether the device works (see [Troubleshooting](https://github.com/linuxwacom/libwacom/wiki/Troubleshooting#my-tablet-doesnt-work))

RayJW:
As discussed in the issue, this adds support for the Lenovo Precision Pen 2 (2023), so it can be properly recognized and remapped by software utilizing libwacom. While not all data is as cut and dry since not everything has been fully documented yet, I think this should be a good starting point, and it seems to be working fine on my side so far.

closes #774


whot:
merged, thaks

Issue #794, opened 2024-10-17
.tablet file for wacom hid 52c6


chrostino:
- **Device name**: Wacom hid 52c6 (from the laptop Lenovo Idepad Flex 5 14ALC7)
- **Device model identifier**: 
- **libwacom version**: 2.2.0-1

- [x] I understand that libwacom does affect whether the device works (see [Troubleshooting](https://github.com/linuxwacom/libwacom/wiki/Troubleshooting#my-tablet-doesnt-work))

- [x] **udevadm info output**: [udevadm info output.txt](https://github.com/user-attachments/files/17409514/udevadm.info.output.txt)

- [x] sysinfo issue in [wacom-hid-descriptors](https://github.com/linuxwacom/wacom-hid-descriptors): https://github.com/linuxwacom/wacom-hid-descriptors/issues/413
 
I run the `sysinfo.sh` script which was unable to generate a `.tablet` file, so I was trying to write one myself manually following instructions. I was able to to get to this point, but I think there are some issues

```
# Wacom
# Wacom HID 52C6 Pen
#
# sysinfo.6QHqFdEfjw
# https://github.com/linuxwacom/wacom-hid-descriptors/issues/413

[Device]
Name=Wacom HID 52C6 Pen
ModelName=
Class=ISDV4
DeviceMatch=i2c:056a:52c6;
Width=11
Height=7
# No pad buttons, so no layout
IntegratedIn=Display;System
Styli=@generic-no-eraser

[Features]
Stylus=true
Touch=true
Buttons=0
```
I am missing the ModelName since it isn't specified how to find it (but as it's said it might not be necessary) but most importantly while testing this `.tablet` file on my system the device is recognized and a new section in the wacom/tablet section of the system settings appears for the tablet section (whereas it was missing before), but the stylus sections still says "no stylus found" so I am unable to change any settings about it (which was my goal to begin with). That's why I concluded that the `.tablet` file I've written is missing something crucial
![image](https://github.com/user-attachments/assets/17537cbd-355a-4217-b9ab-8e398b27e562)
![image](https://github.com/user-attachments/assets/91826255-2700-4614-9ca0-a508957aa16f)

chrostino:
After looking at #774 I believe the main problem described here could be solved by #793.

PR #795, opened 2024-10-17
Create wacom-hid-52c6.tablet

chrostino:
since the `sysinfo.sh` script was unable to generate the `.tablet` file I wrote this manually following instructions and copying from files of similar devices, although I am unsure if the `ModelName=` field needs to be filled (I left it blank since I couldn't find that information) and don't know if the line `Styli=@isdv4-aes` is even necessary. Some of the files of other devices I looked at had it, some didn't.
Also I believed there was a problem in the file I discussed in #794, but closed the issue since I it is fixed by #793

### DIFF
--- a/data/wacom-hid-52c6.tablet
+++ b/data/wacom-hid-52c6.tablet
@@ -1,0 +1,22 @@
+# Wacom
+# Wacom HID 52C6 Pen
+# AES sensor used by the Lenovo Ideapad Flex 5 14ALC7
+#
+# sysinfo.6QHqFdEfjw
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/413
+
+[Device]
+Name=Wacom HID 52C6 Pen,Wacom HID 52C6 Finger
+ModelName=
+Class=ISDV4
+DeviceMatch=i2c:056a:52c6;
+Width=11
+Height=7
+# No pad buttons, so no layout
+IntegratedIn=Display;System
+Styli=@isdv4-aes
+
+[Features]
+Stylus=true
+Touch=true
+Buttons=0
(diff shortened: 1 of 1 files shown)


whot (changes requested):
Thanks, this is mostly good but I'll need you to fix the commit messsage, see e.g. 9a36f7666e04cd49c8ecc438c18840bb04a4743f for the most recent one. It doesn't need to have a lot of info unless there's something really special about the device.

whot:
`ModelName` can be left blank, and for the `Styli` line: run `tools/show-stylus.py` and see what that says. If it doesn't see any stylus IDs you can leave it out, otherwise what you have will be fine.

chrostino:
> `ModelName` can be left blank, and for the `Styli` line: run `tools/show-stylus.py` and see what that says. If it doesn't see any stylus IDs you can leave it out, otherwise what you have will be fine.

it shows the two stylus I have, the lenovo digital pen 2 and the lenovo precision pen 2 (2023), both of which still need to be merged in their pull requests (that I have added to my `/etc/libwacom/wacom.stylus`
Also after terminating in the terminal it shows this
```
Unknown stylus id 0x0. New entry needed
Styli=isdv4-aes
```
should I just put `Styli=@isdv4-aes`?

data/wacom-hid-52c6.tablet
```diff
@@ -0,0 +1,22 @@
...
+# AES sensor used by the Lenovo Ideapad Flex 5 14ALC7
+#
+# sysinfo.6QHqFdEfjw
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/413
+
+[Device]
+Name=Wacom HID 52C6
+ModelName=
+Class=ISDV4
+DeviceMatch=i2c:056a:52c6;
```
whot, line 12:
```suggestion
DeviceMatch=i2c|056a|52c6;
```

whot:
squashed and merged, thanks. Now that there's multiple commits I could even fix the commit message :)

`isdv4-aes` is fine, thanks

PR #796, opened 2024-10-17
Update wacom.stylus - added Lenovo Digital Pen 2

chrostino:
I'm not sure if `BAT_HID` is required. Also as I have understood the lenovo digital pen 2 has tilt "detection", but not tilt "recognition", which I believe means it knows when it's tilted but not how much so I didn't put `Tilt` in the `Axes=` field, but I'm not 100% sure

### DIFF
--- a/data/wacom.stylus
+++ b/data/wacom.stylus
@@ -222,6 +222,15 @@ EraserType=Button
 Axes=Tilt;Pressure
 Type=Mobile
 
+[0x219]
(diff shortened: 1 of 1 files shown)


whot:
if you run `libinput debug-events` or alternatively `libinput record` (kernel level) do you see tilt events with varying (sensible) values?

chrostino:
No it doesn't, so no tilt

whot (approved):

whot:
merged, thanks

PR #805, opened 2024-10-23
Add support for Dial status LEDs

whot:
libinput at least uses the status leds for grouping: our data file says
there's a status led for the ring and the toggle button for the ring is
B, and button are at position P. libinput then uses the fact that there
are two status LEDs to create two groups and matches the buttons
with the position of the mode toggle button (i.e. if button B is in
position P, all buttons in the same position P must belong to the same
mode group).

This requires that the status led for dials exists, so let's add them.

Fixes: https://github.com/linuxwacom/libwacom/commit/b6b82e4afa6a2114afc819cddf6b3ee65dff2e15 ("Add support for relative dials (https://github.com/linuxwacom/libwacom/pull/665)")


whot:
cc @Joshua-Dickens for an extra pair of eyeballs :)

PR #815, opened 2024-10-29
Add Support for Lenovo Active Pen 3 (2023)

TomasVraga:
Hello,

I've encountered the same issue as [#774](https://github.com/linuxwacom/libwacom/issues/774) but with the Active Pen 3.

I've added the stylus into lenovo.stylus and it seems to work fine now.
Here's the results from the meson test log: 
[testlog.txt](https://github.com/user-attachments/files/17561537/testlog.txt)


whot:
thanks!

PR #827, opened 2024-11-07
data: add Huion KeyDial K20

whot:
This is the successor to the KD100 and optically the same so these files are copy/paste from the KD100

@deevad has a nicer-looking svg in [udev-hid-bpf#35](https://gitlab.freedesktop.org/libevdev/udev-hid-bpf/-/issues/35) but unfortunately that one requires a lot of manual fixes for the various CSS classes and object IDs we require. In the interest of getting this done quickly, here's a copy/pasted one that's been modified quickly.

This is a draft because I haven't yet been able to convince gnome settings to show the OSD

### DIFF
--- a/data/huion-keydial-k20.tablet
+++ b/data/huion-keydial-k20.tablet
@@ -1,0 +1,32 @@
+# Huion
+# Keydial K20
+#
+# sysinfo.lvuqy3Kjgl.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/425
+#
+#   __________
+#  |( S )     |
+#  +----------+
+#  |  A B C D |
+#  |  E F G H |
+#  |  I J K L |
+#  |  M N O P |
+#  |   Q  R   |
+#  +----------+
+
+[Device]
+Name=Huion Keydial K20
+ModelName=K20
+# This appears to be a unique PID, if that changes the FW prefix is HUION_T21h
+DeviceMatch=usb|256c|0069
+Layout=huion-keydial-k20.svg
+IntegratedIn=Remote
+
+[Features]
+NumDials=1
+Stylus=false
+DialNumModes=4
+
+[Buttons]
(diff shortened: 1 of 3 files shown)


Deevad:
Thank you very much! 

I also posted [a SVG I made on the udev-ebpf thread](https://gitlab.freedesktop.org/libevdev/udev-hid-bpf/-/issues/35#svg-for-libwacom). Mine is still not cleaned (I can't still run the cleanup python script) but maybe you can enhance the layout you just merged with some element copy/pasted from mine (like the silhouette of the case). 

The SVG: huion-kdial-mini-K20_inkscape-source

And it looks like that: 

huion-kdial-mini-K20_inkscape-screenshot

whot:
Thanks, I do like your SVG better but it needs a fair bit of fixup to pass the test (see #829 to make the tests provide some useful output). I played with it yesterday but ran of out time because there's too much of it... the simple fixes were `s|Ring|Dial|g` to but then all the buttons need fixes. And running `clean_svg.py` on the device moves the mode-switch button around for unknown reasons, it's quite frustrating. Either way, right now I'd prefer we have *something* so we can easily test with the various branch tips instead of having custom branches everywhere.

Deevad:
Yes, I 100% agree. I was just suggesting to copy/paste the case outlines of my SVG on top of your SVG design just for decorating yours. But it is of course super optional.

whot:
> . I was just suggesting to copy/paste the case outlines of my SVG on top of your SVG design just for decorating yours

good idea, done now. This makes it a bit different to the other tablets which do not have an outline but it looks pretty terrible without it so let's see if anyone complains...

Deevad:
> good idea, done now.

Thank you!

whot:
Ok, got this all to work - the dial is missing but that's a factor of the gnome and wayland stack not yet supporting relative dials. So we're good to merge this IMO

Deevad:
> So we're good to merge this IMO

Sure. Thank you for merging it.

Issue #761, opened 2024-09-01
RFC: stylus tool id aliases


whot:
Let's add a line in the form `Aliases=0x1234;0xabcd` to `libwacom.stylus` so we can easily duplicate tools that are identical but have had (or will have) multiple tool ids over time.

Note: this needs to be compatible with the new `vid:pid` notation: #759

whot:
After some private discussion with @Pinglinux this si probably best solved by this approach:
```ini
[0x100802]
Name=Art Pen
PairedStylusIds=0x100804
... other fields...

[0x802]
AliasOf=0x100802
PairedStylusIds=0x804
```
In other words, we still have entries for styli but declare those to be an alias of another stylus and only define the fields where they override the original stylus' setting.


The originally suggested `Aliases=0x802` doesn't work too well for the `PairedStylusIds` settings which also needs an update for aliased styli.

Issue #142, opened 2019-08-05
Use of g_autofree and g_autoptr


whot:
glib has two useful macros `g_autofree` and `g_autoptr` that we could use in our code. They're wrappers around `__attribute__((__cleanup__))` and make things a bit more straightforward:

```
int foo(void) {
   g_autofree char *name = NULL;
   name = strdup("bar");
   return 1;
}
```
The `g_autofree` macro tells the compiler to add the required `g_free(name)` before the return statement. The `g_autoptr` macro is effectively the same but for `GObject`.

Using those would make the code slightly easier in some places, but comes with restrictions:
- they're only supported on gcc, clang and (I think) icc but not MSVC. Not sure we care about that bit anyway.
- our own structs aren't GObjects so we cannot use them, we'd have to define our own cleaup functions. Which is trivial though.

Any comments?

jigpu:
That's a cute trick, but its going to take a while for my inner proofreader to get used to it. Limiting to GCC and Clang should be fine; one of those should be the default compiler on most any Linux or BSD system out there (but maybe(?) not Solaris, if we care :man_shrugging: ).

It looks like the automatic cleanup has [its own footguns](https://blog.fishsoup.net/2015/11/05/attributecleanup-mixed-declarations-and-code-and-goto/) to be avoided, but only their `count1()` example is both buggy and idiomatic. Probably nothing to worry about.

whot:
Simplest rule is "always initialize to NULL" and you're pretty much good to go. Alas, and this is the second time this bites me (we have problems like this in libratbag as well): clang analyzer has random false positives when you start using ``__attribute__((__cleanup__)``. And it does not have a way to shut up specific warnings, so you're left with them until you can fix them.

Here's my branch with the conversions:
https://github.com/whot/libwacom/tree/wip/autoptr

One of the false positives is `bus_str` in `get_device_info()` in `libwacom/libwacom.c`. It took me a lot of trial and error and random guesses to figure out how to make that one disappear. 

The other two were even worse: one was changing from `realloc` to `reallocarray`, the second one from `strdup` to `g_strdup`. Functionally equivalent but one caused false positives, the other one doesn't.

So yeah, the issue with `g_autoptr` and related is that we'll spend more time trying to figure out how to work around clang-analyzer's false positives than we'll save using it in the first place. Right now we're at less than 30 min to switch everything, 3 hours trying things to make c-a shut up.

jigpu:
[...]

PR #844, opened 2024-12-24
Support "AliasOf" keyword

skomra:
Add a new keyword to libwacom.stylus so we can share the settings among those tools that of the same type: RFC: stylus tool id aliases.

https://github.com/linuxwacom/libwacom/issues/761

### DIFF
--- a/data/wacom.stylus
+++ b/data/wacom.stylus
@@ -34,51 +34,26 @@ Type=Mobile
 
 [0x11]
 # Lenovo ; VID_NONE     | 0x0000 | BAT_CHRG
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x19]
 # Lenovo ; VID_LENOVO   | 0x60A8 | BAT_SWAP | BAT_HID
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x21]
 # HP     ; VID_NONE     | 0x0000 | BAT_SWAP
 # Huawei ; VID_NONE     | 0x0000 | BAT_SWAP
 # Lenovo ; VID_NONE     | 0x0000 | BAT_SWAP
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x31]
 # Dell   ; VID_BROADCOM | 0x81B9 | BAT_SWAP
 # HP     ; VID_CHICONY  | 0xB4A3 | BAT_SWAP | BAT_HID
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x39]
 # Huawei ; VID_HUAWEI   | 0x1091 | BAT_SWAP ("Huawei MatePen" / AF61)
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x49]
 # Wacom  ; VID_WACOM    | 0x035F | BAT_SWAP | BAT_GATT | BAT_SHARED | LONGPRESS ("Wacom Bamboo Ink" / CS321A)
@@ -91,145 +66,80 @@ Type=Mobile
 
 [0x71]
 # Wacom  ; VID_WACOM    | 0x035F | BAT_SWAP | BAT_GATT | BAT_SHARED | LONGPRESS ("Wacom Bamboo Ink" / CS321A1)
-Name=Bamboo Ink
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x49
 
 [0x221]
 # HP     ; VID_CHICONY  | 0xB4A3 | BAT_SWAP | BAT_HID
 # Lenovo ; VID_NONE     | 0x0000 | BAT_SWAP
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x231]
 # Dell   ; VID_BROADCOM | 0x81C6 | BAT_SWAP | LONGPRESS (Dell PN557W)
 # HP     ; VID_CHICONY  | 0xB4A3 | BAT_SWAP | BAT_HID
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x271]
 # Wacom  ; VID_NONE     | 0x0000 | BAT_SWAP ("Wacom Bamboo Ink" / CS323A)
-Name=Bamboo Ink
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x49
 
 [0x421]
 # HP     ; VID_NONE     | 0x0000 | BAT_SWAP
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x431]
 # Dell   ; VID_BROADCOM | 0x81B9 | BAT_SWAP
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x621]
 # Lenovo ; VID_NONE     | 0x0000 | BAT_SWAP
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x631]
 # Dell   ; VID_NONE     | 0x0000 | BAT_SWAP | LONGPRESS (Dell PN557W)
 # Dell   ; VID_NONE     | 0xf142541e | BAT_SWAP (Dell PN5122W)
 Name=Dell Active Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x8051]
 # Google ; VID_NONE     | 0x0000 | BAT_SWAP ("Google Pixelbook Pen" / C0B)
-Name=AES Pen
-Group=isdv4-aes
 Buttons=0
-EraserType=Button
 Axes=Tilt;Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x805B]
 # Dell   ; VID_BROADCOM | 0x81D5 | BAT_SWAP | BAT_GATT | BAT_SHARED | LONGPRESS (Dell PN579X)
 # Lenovo ; VID_LENOVO   | 0x60C5 | BAT_SWAP | BAT_GATT | BAT_SHARED
 # Toshiba; VID_NONE     | 0x0000 | BAT_SWAP
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
 Axes=Tilt;Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x806B]
 # HP     ; VID_CHICONY  | 0x1728 | BAT_CHRG | BAT_PROX | BAT_SHARED | LONGPRESS | PROX ("HP Rechargeable Active Pen" / HP Active Pen G2 / 4KL69AA)
 # Huawei ; VID_NONE     | 0x0000 | BAT_SWAP
 # Lenovo ; VID_LENOVO   | 0x60C2 | BAT_CHRG | BAT_GATT | BAT_SHARED
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
 Axes=Tilt;Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x807B]
 # Wacom  ; VID_WACOM    | 0x0397 | BAT_CHRG | BAT_GATT | BAT_SHARED | LONGPRESS ("Wacom Bamboo Ink Plus" / CS322A)
 Name=Bamboo Ink Plus
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
+AliasOf=0x49
 Axes=Tilt;Pressure
-Type=Mobile
 
 [0x826B]
 # HP     ; VID_CHICONY  | 0x1728 | BAT_CHRG | BAT_PROX | BAT_SHARED | LONGPRESS | PROX ("HP Active Pen G3" / L08263-003 / L57042-001)
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
 Axes=Tilt;Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x846B]
 # HP     ; VID_CHICONY  | 0x1850 | BAT_CHRG | BAT_GATT | BAT_SHARED | LONGPRESS | PROX ("HP Rechargeable Active Pen G3" / HP Active Pen G3 / 6SG43AA)
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
 Axes=Tilt;Pressure
-Type=Mobile
+AliasOf=0x1
 
 [0x219]
 # Lenovo ; VID_LENOVO     | 0x219 | BAT_SWAP
 Name=Lenovo Digital Pen 2
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
 
 # Inking pen have no eraser
 [0x812]
--- a/libwacom/libwacom-database.c
+++ b/libwacom/libwacom-database.c
@@ -47,6 +47,11 @@
 #define BUTTONS_GROUP "Buttons"
 #define KEYS_GROUP "Keys"
 
+typedef enum {
+        IGNORE_ALIASES  = 0,
+        ONLY_ALIASES    = 1,
+} AliasStatus;
+
 static WacomClass
 libwacom_class_string_to_enum(const char *class)
 {
@@ -105,6 +110,34 @@ type_from_str (const char *type)
 	return WSTYLUS_UNKNOWN;
 }
 
+static const char *
+str_from_type (WacomStylusType type)
+{
+	switch (type) {
+	case WSTYLUS_UNKNOWN:
+		return NULL;
+	case WSTYLUS_GENERAL:
+		return "General";
+	case WSTYLUS_INKING:
+		return "Inking";
+	case WSTYLUS_AIRBRUSH:
+		return "Airbrush";
+	case WSTYLUS_CLASSIC:
+		return "Classic";
+	case WSTYLUS_MARKER:
+		return "Marker";
+	case WSTYLUS_STROKE:
+		return "Stroke";
+	case WSTYLUS_PUCK:
+		return "Pick";
+	case WSTYLUS_3D:
+		return "3D";
+	case WSTYLUS_MOBILE:
+		return "Mobile";
+	}
+	return NULL;
+}
+
 static WacomEraserType
 eraser_type_from_str (const char *type)
 {
@@ -117,6 +150,22 @@ eraser_type_from_str (const char *type)
 	if (g_str_equal(type, "Button"))
 		return WACOM_ERASER_BUTTON;
 	return WACOM_ERASER_UNKNOWN;
+}
+
+static const char *
+eraser_str_from_type (WacomEraserType type)
+{
+	switch (type) {
+	case WACOM_ERASER_NONE:
+		return "None";
+	case WACOM_ERASER_INVERT:
+		return "Invert";
+	case WACOM_ERASER_BUTTON:
+		return "Button";
+	case WACOM_ERASER_UNKNOWN:
+		return NULL;
+	}
+	return NULL;
 }
 
 WacomBusType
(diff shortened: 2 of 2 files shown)


libwacom/libwacom-database.c
```diff
@@ -279,8 +323,88 @@ parse_stylus_id(const char *str, WacomStylusId *id)
 	return ret;
 }
 
+static char *
+string_or_fallback (GKeyFile *keyfile, const char *group,
+		    const char *key, const char *fallback) {
```
whot, line 328:
```suggestion
		    const char *key, const char *fallback) 
		    {
```
including the others below too please

libwacom/libwacom-database.c
```diff
@@ -279,8 +323,88 @@ parse_stylus_id(const char *str, WacomStylusId *id)
...
+	g_autoptr(GError) local_error = NULL;
+	if (g_key_file_has_key(keyfile, group, key, &local_error))
+		return g_key_file_get_integer(keyfile, group, key, NULL);
+
+	return fallback;
+}
+
+static bool
+boolean_or_fallback (GKeyFile *keyfile, const char *group,
+		 const char *key, bool fallback, WacomStylus *stylus) {
```
whot, line 348: looks like indentation is off here and in the `int_or_fallback` too

libwacom/libwacom-database.c
```diff
@@ -279,8 +323,88 @@ parse_stylus_id(const char *str, WacomStylusId *id)
...
+{
+	gchar **values;
+	WacomStylusId *id;
+	char buffer[33];
+	bool ret;
+
+	if (array == NULL)
+		return NULL;
+
+	values = g_new (gchar *, array->len + 1);
```
whot, line 370:
```suggestion
	values = g_new0 (gchar *, array->len + 1);
```
saves you manually null-terminating it a few lines below. Having said that - we're already using glib here so this could probably be something like:
```c

g_autoptr(GArray) *a = g_array_new(true, false, array->len);
for (...) {
   g_array_append_vals(...)
}
return g_array_steal(a);
```

libwacom/libwacom-database.c
```diff
@@ -279,8 +323,88 @@ parse_stylus_id(const char *str, WacomStylusId *id)
...
+	const char *vidstr;
+	int vid;
+	bool ret = false;
+	int aliased_int;
+
+        if (tokens[1] == NULL) {
+                vidstr = G_STRINGIFY(WACOM_VENDOR_ID);
+        } else {
+                vidstr = tokens[0];
+        }
```
whot, line 394: up to here we have the same code from `parse_stylus_id`, we should be able to use that and then override `id->tool_id` with just the aliased int. I think...

better than having the parsing code duplicated.

libwacom/libwacom-database.c
```diff
@@ -295,32 +419,65 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
-		stylus->name = g_key_file_get_string(keyfile, groups[i], "Name", NULL);
-		stylus->group = g_key_file_get_string(keyfile, groups[i], "Group", NULL);
+
+		if (alias == IGNORE_ALIASES && g_key_file_get_string(keyfile, groups[i], "AliasOf", NULL) != NULL) {
+			continue;
+		}
+
+		if (alias == ONLY_ALIASES && g_key_file_get_string(keyfile, groups[i], "AliasOf", NULL) == NULL) {
+			continue;
+		}
```
whot, line 444:
```suggestion
		g_autofree char *aliasstr = g_key_file_get_string(keyfile, groups[i], "AliasOf", NULL);
		if (alias == IGNORE_ALIASES && aliasstr) {
			continue;
		}

		if (alias == ONLY_ALIASES && !aliasstr) {
			continue;
		}
```
because there's no point doing `g_key_file_get_string()` three times for an alias

libwacom/libwacom-database.c
```diff
@@ -279,8 +323,88 @@ parse_stylus_id(const char *str, WacomStylusId *id)
...
+                ret = true;
+        }
+        g_clear_pointer(&tokens, g_strfreev);
+
+        return ret;
+}
+
 static void
-libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
+libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path, AliasStatus alias)
```
whot, line 407: nitpick:

```suggestion
libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path, AliasStatus handle_aliases)
```
might make this more immediately understandable (and lets you use `alias` for the actual alias)

libwacom/libwacom-database.c
```diff
@@ -295,32 +419,65 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
+		}
+
+		if (alias == ONLY_ALIASES && g_key_file_get_string(keyfile, groups[i], "AliasOf", NULL) == NULL) {
+			continue;
+		}
+
+		if (alias == ONLY_ALIASES) {
+			alias_value = g_key_file_get_string(keyfile, groups[i], "AliasOf", NULL);
+
+			if (!create_stylus_id_from_aliased(groups[i], alias_value, &lookup)) {
```
whot, line 449: nickpick on the name:
```suggestion
			if (!create_stylus_id_from_alias(groups[i], alias_value, &lookup)) {
```

libwacom/libwacom-database.c
```diff
@@ -295,32 +419,65 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
 
 		/* We have to keep the integer array for libwacom_get_supported_styli() */
 		stylus->deprecated_paired_ids = g_array_new (FALSE, FALSE, sizeof(int));
 		stylus->paired_styli = g_array_new (FALSE, FALSE, sizeof(WacomStylus*));
 
-		string_list = g_key_file_get_string_list (keyfile, groups[i], "PairedStylusIds", NULL, NULL);
+		if (alias == IGNORE_ALIASES) {
+			string_list = g_key_file_get_string_list(keyfile, groups[i], "PairedStylusIds", NULL, NULL);
+		} else {
+			string_list = g_key_file_get_string_list(keyfile, groups[i], "PairedStylusIds", NULL, NULL);
```
whot, line 475:
```suggestion
		string_list = g_key_file_get_string_list(keyfile, groups[i], "PairedStylusIds", NULL, NULL);
		if (alias != IGNORE_ALIASES) {
```
does the same thing I think

libwacom/libwacom-database.c
```diff
@@ -333,47 +490,47 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
-		stylus->has_lens = g_key_file_get_boolean(keyfile, groups[i], "HasLens", &error);
-		if (error && error->code == G_KEY_FILE_ERROR_INVALID_VALUE)
-			g_warning ("Stylus %s (%s) %s\n", stylus->name, groups[i], error->message);
-		g_clear_error (&error);
-		stylus->has_wheel = g_key_file_get_boolean(keyfile, groups[i], "HasWheel", &error);
-		if (error && error->code == G_KEY_FILE_ERROR_INVALID_VALUE)
-			g_warning ("Stylus %s (%s) %s\n", stylus->name, groups[i], error->message);
-		g_clear_error (&error);
+		stylus->has_lens = boolean_or_fallback(keyfile, groups[i], "HasLens", other ? other->has_lens : FALSE, stylus);
+
```
whot, line 494:
```suggestion
```

libwacom/libwacom-database.c
```diff
@@ -1093,12 +1250,11 @@ load_stylus_files(WacomDeviceDatabase *db, const char *datadir)
 
 	while ((file = readdir(dir))) {
 		char *path;
-
 		if (!is_stylus_file(file))
 			continue;
 
 		path = g_build_filename (datadir, file->d_name, NULL);
-		libwacom_parse_stylus_keyfile(db, path);
+		libwacom_parse_stylus_keyfile(db, path, alias);
```
whot, line 1257: heh, see my comment above for immediately readable by naming to `handle_aliases` because I looked at this and was wondering why we're passing an alias into the keyfile parser

libwacom/util-strings.h
```diff
@@ -132,6 +132,21 @@ safe_atoi_base(const char *str, int *val, int base)
...
+
+	if (base == 10)
+		snprintf(buffer, buf_size, "0x%08d", val);
+	else if (base == 16)
+		snprintf(buffer, buf_size, "0x%08x", val);
+	else if (base == 8)
+		snprintf(buffer, buf_size, "0x%08o", val);
+
+	return true;
+}
```
whot, line 148: i know this feels pointless but can we add some test cases for this so we never have to worry about this again?

Arguably a `switch` on the base could look nicer but I'm not going to die on that particular hill :)

libwacom/libwacom-database.c
```diff
@@ -279,8 +323,88 @@ parse_stylus_id(const char *str, WacomStylusId *id)
...
+	bool ret;
+
+	if (array == NULL)
+		return NULL;
+
+	values = g_new (gchar *, array->len + 1);
+	for (guint i = 0; i < array->len; i++) {
+		id = &g_array_index(array, WacomStylusId, i);
+		ret = safe_itoa_base(id->tool_id, buffer, sizeof(buffer), 16);
+		values[i] = strdup(buffer);
```
whot, line 374: `safe_itoa_base` always returns true so the `ret` assignment is pointless (and we don't check it here anyway). But since this is appears to be the only place we need it we can skip that helper altogether and make it a:

```suggestion
		values[i] = g_strdup_printf("0x%08x", id->tool_id);
```

libwacom/libwacom-database.c
```diff
@@ -295,32 +419,65 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
 
 	groups = g_key_file_get_groups (keyfile, NULL);
 	for (i = 0; groups[i]; i++) {
-		WacomStylus *stylus;
-		WacomStylusId id;
+		WacomStylus *stylus, *other;
+		WacomStylusId id, lookup;
```
whot, line 423: Move `lookup` into the condition where we use it please, makes the code more obvious.

libwacom/libwacom-database.c
```diff
@@ -295,32 +419,65 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
 
 	groups = g_key_file_get_groups (keyfile, NULL);
 	for (i = 0; groups[i]; i++) {
-		WacomStylus *stylus;
-		WacomStylusId id;
+		WacomStylus *stylus, *other;
```
whot, line 422: Is there a word like `aliasee`? Or maybe `source_stylus`? Or some other word that tells us immediately that this is the one we're aliasing to (from?).

libwacom/libwacom-database.c
```diff
@@ -279,8 +323,88 @@ parse_stylus_id(const char *str, WacomStylusId *id)
 	return ret;
 }
 
+static char *
+string_or_fallback (GKeyFile *keyfile, const char *group,
+		    const char *key, const char *fallback) {
+	g_autoptr(GError) local_error = NULL;
```
whot, line 329: Regarding the use of `g_autoptr` see my comment in https://github.com/linuxwacom/libwacom/issues/142#issuecomment-2290465558 - I think right now scan-build is more useful to us than g_autoptr/g_autofree so as annoying as it is we should do these manually.

I've used it in my suggestions below but only because it's easier to type out here than adding the correct free lines, so please adjust for that

data/wacom.stylus
```diff
@@ -34,51 +34,26 @@ Type=Mobile
 
 [0x11]
 # Lenovo ; VID_NONE     | 0x0000 | BAT_CHRG
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
+AliasOf=0x1
```
whot, line 37: ack to all these since they seem to be the same hw

data/wacom.stylus
```diff
@@ -91,162 +66,89 @@ Type=Mobile
...
 
 # Inking pen have no eraser
 [0x812]
 # Intuos and Intuos2
-Name=Inking Pen
+AliasOf=0x20802
```
whot, line 145: Is this really worth changing these? The I2 is now old enough that we should never have to think about it again so leaving this code untouched forever is better. This applies to a whole bunch of others too, IMO using aliases is best for where it's really "same thing, different id" even though the alias implementation lets us almost use it like CSS.

whot (changes requested):
Thanks for the patches - overall looks good with a few minor code changes. As for the actual stylus file: I think we should be minimally aggressive on using the aliases, especially for old and discontinued hw where it's just easier to not ever touch this again.

libwacom/libwacom-database.c
```diff
@@ -279,8 +323,81 @@ parse_stylus_id(const char *str, WacomStylusId *id)
...
+	if (array == NULL)
+		return NULL;
+
+	GArray *a = g_array_new(true, false, array->len);
+	for (guint i = 0; i < array->len; i++) {
+		id = &g_array_index(array, WacomStylusId, i);
+		g_array_append_vals(a, g_strdup_printf("0x%08x", id->tool_id), 1);
+	}
+
+	return g_array_steal(a, (long unsigned int *)(&(a->len)));
```
whot, line 396: That cast can have interesting side-effects if `sizeof(a->len) != sizeof(long unsigned int)`, an easier way around this is:

```suggestion
	gsize len = a->len;
	return g_array_steal(a, len);
```

libwacom/libwacom-database.c
```diff
@@ -279,8 +323,81 @@ parse_stylus_id(const char *str, WacomStylusId *id)
...
+	}
+
+	if (ret)
+		return g_key_file_get_boolean(keyfile, group, key, &error);
+
+	return fallback;
+}
+
+static gchar **
+garray_to_string_list(GArray *array)
```
whot, line 383: this sounds like it's a generic array->string list conversion but it's a `WacomStylusId` -> hex string converstion, let's rename this please, e.g. 

```
stylus_ids_as_hex(GArray *array)
```
or something similar

libwacom/libwacom-database.c
skomra: That is giving me the following error. I'm not sure why g_array_steal wants a pointer. 
```
../libwacom/libwacom-database.c: In function ‘stylus_ids_as_hex’:
../libwacom/libwacom-database.c:397:33: error: passing argument 2 of ‘g_array_steal’ makes pointer from integer without a cast [-Wint-conversion]
  397 |         return g_array_steal(a, len);
      |                                 ^~~
      |                                 |
      |                                 gsize {aka long unsigned int}
In file included from /usr/include/glib-2.0/glib.h:33,
                 from ../libwacom/libwacomint.h:34,
                 from ../libwacom/libwacom-database.c:30:
/usr/include/glib-2.0/glib/garray.h:86:54: note: expected ‘gsize *’ {aka ‘long unsigned int *’} but argument is of type ‘gsize’ {aka ‘long unsigned int’}
   86 |                                    gsize            *len);
      |                                    ~~~~~~~~~~~~~~~~~~^~~
```

whot: Edited my suggestion above  - you can pass `&len` instead but after checking the source for `g_array_steal()` - len is an output-only argument and returns the number of elements in the array, we can pass NULL since we don't care about that value here.

libwacom/libwacom-database.c
```diff
@@ -279,8 +323,81 @@ parse_stylus_id(const char *str, WacomStylusId *id)
...
+{
+	WacomStylusId *id;
+
+	if (array == NULL)
+		return NULL;
```
whot, line 388:
```suggestion
	if (array == NULL || array->len == 0)
		return NULL;
```
to avoid
```
not ok /load/0000:0000 - GLib-FATAL-CRITICAL: g_array_new: assertion 'elt_size > 0' failed
```

libwacom/libwacom-database.c
```diff
@@ -295,32 +412,69 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
 			continue;
 		}
+
 		stylus = g_new0 (WacomStylus, 1);
 		stylus->refcnt = 1;
 		stylus->id = id;
-		stylus->name = g_key_file_get_string(keyfile, groups[i], "Name", NULL);
-		stylus->group = g_key_file_get_string(keyfile, groups[i], "Group", NULL);
+
+		char *aliasstr = g_key_file_get_string(keyfile, groups[i], "AliasOf", NULL);
```
whot, line 431: Note that `aliasstr` leaks in every `continue` case below (except the `!aliasstr` one)

libwacom/libwacom-database.c
```diff
@@ -295,32 +412,69 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
+		int aliased_int;
 
 		if (!parse_stylus_id(groups[i], &id)) {
 			g_warning ("Failed to parse stylus ID '%s', ignoring entry", groups[i]);
 			continue;
 		}
+
 		stylus = g_new0 (WacomStylus, 1);
 		stylus->refcnt = 1;
 		stylus->id = id;
```
whot, line 429: You're allocating `stylus` here but it's not used until after the `g_clear_pointer(aliasstr)` line - any `continue` between here and later will leak `stylus` - IOW move those three lines down past the aliasstr handling (or rather, don't add all the new code above `g_new0` :)

skomra:
I needed to run the full check on the code changes so I pushed the first patch. I'm still working on the updated data file.

whot:
> I needed to run the full check on the code changes so I pushed the first patch. I'm still working on the updated data file.

what I sometimes do is to `git switch -c somebranch-v2`, then work on that and push it (without filing an MR but looking at the `Actions` in github) and when it finally passes I move the patches over to the `somebranch` (cherry-pick or git reset) and push that to the already-open PR. This way it's less embarrassing for me when I have to play trial-and-error with the CI because no-one sees the pushes :)

skomra:
Thanks for the tip! I think I used to do that but I haven't been active on github in a while and had forgotten that.

The data file patch has been added and the PR is ready for review again. Thanks for your help.

libwacom/libwacom-database.c
```diff
@@ -295,32 +412,71 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
+
+		char *aliasstr = g_key_file_get_string(keyfile, groups[i], "AliasOf", NULL);
+
+		if (handle_aliases == IGNORE_ALIASES && aliasstr) {
+			g_clear_pointer(&aliasstr, g_free);
+			continue;
+		}
+
+		if (handle_aliases == ONLY_ALIASES && !aliasstr) {
+			g_clear_pointer(&aliasstr, g_free);
```
whot, line 435: not really necessary since we explicitly check for it to be NULL but it doesn't hurt and makes it more symmetric I guess.

libwacom/libwacom-database.c
```diff
@@ -295,32 +412,71 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
+				g_warning ("Failed to parse Aliased ID '%s', ignoring entry", groups[i]);
+				continue;
+			}
+			if (safe_atoi_base (aliasstr, &aliased_int, 16))
+				lookup.tool_id = aliased_int;
+
+			aliased = g_hash_table_lookup(db->stylus_ht, &lookup);
+			if (!aliased) {
+				g_warning ("Unkown Aliased ID 0x%08x %s\n", lookup.tool_id, aliasstr);
+				continue;
```
whot, line 456: both of the `continue`s here will leak `stylus`.

You could trigger this with a crafted `StylusEntry` in `test_libwacom.py` but since we don't run those tests in valgrind.... :shrug:

libwacom/libwacom-database.c
```diff
@@ -295,32 +412,71 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
+			}
+		}
+		g_clear_pointer(&aliasstr, g_free);
+
+		stylus->name = string_or_fallback(keyfile, groups[i], "Name", aliased ? aliased->name : NULL, NULL);
```
whot, line 461: If you move the `stylus = g_new0` lines to here you don't need to worry about a leaking `stylus` above

libwacom/libwacom-database.c
```diff
@@ -295,32 +412,71 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
 		stylus->eraser_type = eraser_type_from_str (eraser_type);
 		g_clear_pointer(&eraser_type, g_free);
 
 		/* We have to keep the integer array for libwacom_get_supported_styli() */
 		stylus->deprecated_paired_ids = g_array_new (FALSE, FALSE, sizeof(int));
 		stylus->paired_styli = g_array_new (FALSE, FALSE, sizeof(WacomStylus*));
 
-		string_list = g_key_file_get_string_list (keyfile, groups[i], "PairedStylusIds", NULL, NULL);
+		string_list = g_key_file_get_string_list(keyfile, groups[i], "PairedStylusIds", NULL, NULL);
+		if (handle_aliases != IGNORE_ALIASES) {
```
whot, line 474:
```suggestion
		if (handle_aliases == ONLY_ALIASES) {
```
we only have two values, might as well check for the explicit one

libwacom/libwacom-database.c
```diff
@@ -333,47 +489,51 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
 			g_warning ("Stylus %s (%s) %s\n", stylus->name, groups[i], error->message);
 		g_clear_error (&error);
-		stylus->has_wheel = g_key_file_get_boolean(keyfile, groups[i], "HasWheel", &error);
+		stylus->has_wheel = boolean_or_fallback(keyfile, groups[i], "HasWheel", aliased ? aliased->has_wheel : FALSE, stylus, error);
 		if (error && error->code == G_KEY_FILE_ERROR_INVALID_VALUE)
 			g_warning ("Stylus %s (%s) %s\n", stylus->name, groups[i], error->message);
 		g_clear_error (&error);
-
-		stylus->num_buttons = g_key_file_get_integer(keyfile, groups[i], "Buttons", &error);
+		stylus->num_buttons = int_or_fallback(keyfile, groups[i], "Buttons", aliased ? aliased->num_buttons : 0, error);
```
whot, line 500: I keep thinking that for the alias is null case:
```
WacomStylus empty = {0};
WacomStylus *aliased = &empty;
```
then you don't need all the `aliased ? aliased->foo : 0` conditions.

libwacom/libwacom-database.c
```diff
@@ -333,47 +489,51 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
-				flag = WACOM_AXIS_TYPE_DISTANCE;
-			} else if (g_str_equal(string_list[j], "Pressure")) {
-				flag = WACOM_AXIS_TYPE_PRESSURE;
-			} else if (g_str_equal(string_list[j], "Slider")) {
-				flag = WACOM_AXIS_TYPE_SLIDER;
-			} else {
-				g_warning ("Invalid axis %s for stylus ID %s\n",
-					   string_list[j], groups[i]);
+
+		if (handle_aliases == ONLY_ALIASES && (string_list == NULL)) {
```
whot, line 509:
```suggestion
		if (handle_aliases == ONLY_ALIASES && string_list == NULL) {
```

libwacom/libwacom-database.c
```diff
@@ -333,47 +489,51 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
+				} else if (g_str_equal(string_list[j], "Distance")) {
+					flag = WACOM_AXIS_TYPE_DISTANCE;
+				} else if (g_str_equal(string_list[j], "Pressure")) {
+					flag = WACOM_AXIS_TYPE_PRESSURE;
+				} else if (g_str_equal(string_list[j], "Slider")) {
+					flag = WACOM_AXIS_TYPE_SLIDER;
+				} else {
+					g_warning ("Invalid axis %s for stylus ID %s\n",
+						   string_list[j], groups[i]);
+			}
```
whot, line 527:
```suggestion
				}
```

libwacom/libwacom.h
```diff
@@ -309,6 +309,10 @@ typedef enum {
 	WACOM_STATUS_LED_DIAL2		= 5,
 } WacomStatusLEDs;
 
+typedef enum {
+	IGNORE_ALIASES	= 0,
+	ONLY_ALIASES	= 1,
+} AliasStatus;
```
whot, line 315: This one must not be in `libwacom.h`, otherwise it becomes public API. Just putting it into `libwacom-database.c` is enough, it's not used anywhere else.

libwacom/libwacom-database.c
```diff
@@ -279,8 +328,81 @@ parse_stylus_id(const char *str, WacomStylusId *id)
...
+{
+	WacomStylusId *id;
+
+	if (array == NULL || array->len == 0)
+		return NULL;
+
+	GArray *a = g_array_new(true, false, array->len);
+	for (guint i = 0; i < array->len; i++) {
+		id = &g_array_index(array, WacomStylusId, i);
+		g_array_append_vals(a, g_strdup_printf("0x%08x", id->tool_id), 1);
```
whot, line 398: see related comment below: best to append the vendor id here too to make this a full stylus id in the form `0x1234:0xabcd`

whot (changes requested):
a few nitpicks regarding style but the biggest issue here is that there's still a memleak on some of the error paths

libwacom/libwacom-database.c
```diff
@@ -295,32 +417,73 @@ libwacom_parse_stylus_keyfile(WacomDeviceDatabase *db, const char *path)
...
+		if (handle_aliases == ONLY_ALIASES && !aliasstr) {
+			g_clear_pointer(&aliasstr, g_free);
+			continue;
+		}
+
+		if (handle_aliases == ONLY_ALIASES) {
+			WacomStylusId lookup;
+
+			if (!parse_stylus_id(groups[i], &lookup)) {
+				g_warning ("Failed to parse Aliased ID '%s', ignoring entry", groups[i]);
```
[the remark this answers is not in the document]
whot: staring at this again: this check is superflous since we unconditionally check the parsing of `groups[i]` about 20 lines north of here anyway.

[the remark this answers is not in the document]
skomra: Hi Peter, You mention above that "there's still a memleak on some of the error paths" (plural) but I think line 454 is the only memleak you mention in your comments here. Is there a second memleak besides this one?

libwacom/libwacom-database.c
skomra: The way this appeared, it is your last comment. So by below I think you're referring to the comment right above this, just double checking that I understand you right.

skomra: I just noticed #871 - so you can ignore the comments I just made. Sorry about that I was going through my mail oldest to newest.

whot:
Superseded by #871

PR #845, opened 2024-12-29
data: add XP Pen Innovator 16 definition & layout

razcore-rad:
I'm on Ubuntu 24.10 Wayland, am I misunderstanding something here? Weren't there supposed to be options to configure the buttons, dials etc in the options? Thanks
image


whot:
thanks, merged

PR #856, opened 2025-02-13
data: Add 3rd gen Intuos Pro devices

jigpu:


jigpu:
I've added sysinfo output for the small and medium versions of this device. Unfortunately, I do not have a copy of the large version available.

whot:
That CI check only does a `grep sysinfo` so easiest workaround here is putting something like "sysinfo not available, if you have this device please file an issue here..." or something. We can outsmart the computer that way ;)

jigpu:
A copy of the sysinfo for the Large device has been obtained and uploaded

whot (approved):

whot:
> A copy of the sysinfo for the Large device has been obtained and uploaded

nice, thanks!

Issue #859, opened 2025-02-23
github main.yml -- workaround still required?


AesaraB:
Hi, the issue mentioned in the following code has been closed. Is the workaround still required?

https://github.com/linuxwacom/libwacom/blob/6ae945fdef3d06dcb189c82909edf702aef627a5/.github/workflows/main.yml#L21-L25

whot:
probably not if it's been fixed :) Can you send a PR please removing this, then we'll know (if the CI passes, that is).

PR #860, opened 2025-02-23
github: Remove redundant meson build option

AesaraB:
As per #859

Peter wanted to see if the CI fails after excising some code.

However, I am not very familiar with GitHub CI, so I can't say whether the CI will use the clang compiler (for which the workaround was provisioned).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,11 +18,6 @@ jobs:
         compiler:
           - gcc
           - clang
-        meson_options:
-          - ''
-          # clang requires b_lundef=false for b_santize, see
-          # https://github.com/mesonbuild/meson/issues/764
-          - '-Db_sanitize=address,undefined -Db_lundef=false'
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -19,10 +19,9 @@ jobs:
           - gcc
           - clang
         meson_options:
+          # Build options per compiler here
+          - ''
```
whot, line 23: i think we can just drop the `meson_options` altogether - whoever has to add them in the future can look up the git history or figure it out anew :)

AesaraB:
Tests passed, ready for review

AesaraB:
Tests are using the appropriate compiler still

AesaraB:
I may however change the name of the commit to reflect the removal of the compiler options altogether

whot (approved):

whot:
thanks!

PR #865, opened 2025-03-05
Add the XP Pen Deco01 v3

whot:
Basically the same as the v2, at least as seen by libwacom (bar name and pid).

cc @Deevad - can you submit the sysinfo to [wacom-hid-descriptors](https://github.com/linuxwacom/wacom-hid-descriptors) please, this requires the physical so I can't do those with the hid-recordings from https://gitlab.freedesktop.org/libevdev/udev-hid-bpf/-/issues/54

### DIFF
--- a/data/xp-pen-deco01-v3.tablet
+++ b/data/xp-pen-deco01-v3.tablet
@@ -1,0 +1,29 @@
+# XP-Pen
+# Deco01 v3
+#
+# sysinfo.uWI7cjQylf.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/468
+
+[Device]
+Name=XP-Pen Deco 01 V3
+ModelName=
+DeviceMatch=usb|28bd|0947
+PairedIDs=
+Width=10
+Height=6
+IntegratedIn=
+# Looks the same as the v2, so let's re-use the layout
+Layout=xp-pen-deco01-v2.svg
+Styli=@generic-no-eraser
+
+[Features]
+Stylus=true
+Reversible=true
+Touch=false
+TouchSwitch=false
+NumRings=0
+NumStrips=0
+
+[Buttons]
+Left=A;B;C;D;E;F;G;H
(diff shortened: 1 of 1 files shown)


Deevad:
Thank you, I added the sysinfo.

I tried on my Fedora GNOME test laptop to copy manually the .tablet file (and ensure my version had the 01v2.svg layout), but all I have is this: 

Screenshot From 2025-03-07 00-24-04

And the "Map Buttons" button does nothing on click. What did I missed? 

Also, I confirm on this Fedora 41 with GNOME 47.4 that the default layout for the three top button is also 
- button 1: left mouse click
- button 2: middle mouse click
- button 3: right mouse click
- and button 4 to 8 inactive.

So, I can't blame my Debian X11 Plasma system for this behavior. It looks like the pad has this default after installing the udev-hid-rules, no matter the GNU/Linux D.E.

whot:
> And the "Map Buttons" button does nothing on click. What did I missed?

xorg or wayland? tbh all my tests here were on wayland, I'm wondering if there's some bug with the xorg impl.

Fixed the name and added the sysinfo, thanks for filing that.

Deevad:
> xorg or wayland? 

Thanks for the clue. Indeed, I selected the "GNOME" session on the login screen and I assumed this was Wayland by default and it wasn't. I took the session GNOME(Wayland), and this time it did work. Thanks.

I updated the screenshot and the method on the blog-post! https://www.davidrevoy.com/article1066/xppen-deco-01v3-review-on-gnulinux , That's a much better support now on the GNOME side, thanks!

whot:
If you find the time to file a bug for mutter/settings I *might* find the time fixing it for xorg but that's a bit lower on my priority unfortunately. Thanks for confirming though, merged now.

Deevad:
> If you find the time to file a bug for mutter/settings I might find the time fixing it for xorg

No need to report it, it's ok imo to let this one as it is and just inform the Wayland session works. I don't think any GNOME user serious about digital painting keep using recent GNOME version with X11. Especially if it is already like that on Fedora main release and passed all beta tests. I guess the removal of the X11 session is for soon in any case.

PR #867, opened 2025-03-17
data: prefix all styli with a vendor ID

whot:


whot:
cc @skomra since this overlaps with #844

PR #871, opened 2025-03-18
Support "AliasOf" keyword 

### LINKED ISSUE #761: RFC: stylus tool id aliases

whot:
Add a new keyword to libwacom.stylus so we can share the settings among those tools that of the same type.

This supersedes #844 and has the various missing bits:
- style changes and memleak fixes
- better error path handling through `g_set_error()`
- support for full stylus ids (`vid:pid` format)
- test case added so we can make sure the overrides actually works

Closes #761

Closes #884

### DIFF
--- a/data/wacom.stylus
+++ b/data/wacom.stylus
@@ -34,51 +34,26 @@ Type=Mobile
 
 [0x56a:0x11]
 # Lenovo ; VID_NONE     | 0x0000 | BAT_CHRG
-Name=AES Pen
-Group=isdv4-aes
-Buttons=1
-EraserType=Button
-Axes=Pressure
-Type=Mobile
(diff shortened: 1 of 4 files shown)


skomra (approved):
Reviewed-by: Erin Skomra <erin.skomra@wacom.com>

whot:
Added another test for invalid aliases which showed up a missing `continue` in an error case. Fixed now so this should be ready to go.

PR #873, opened 2025-03-20
Support for XP-Pen Deco Pro S

lukasz-tuz:
Pull request adds support for XP-Pen Deco Pro S tablet. This model is virtually identical to the XP-Pen Deco Pro SW tablet, just without the Bluetooth connectivity.


whot:
thanks!

PR #885, opened 2025-06-13
Fix segfault after running list-local-devices

frozolotl:
The regression was caused by ac375537f2592307f6056dbd01f8321eef8af1f6.

The command appears do what it needs to do, it simply segfaults after everything was executed.


whot:
whoah, oops. sorry about that

gayeagle:
Thanks for posting this, I was about to open an issue!

Issue #864, opened 2025-03-03
Switch Width/Height to mm


whot:
IIRC, originally we added w/h to make it possible to refer to devices as "That's the 6x9 intuos 3". Since Wacom uses inches, the width/height is specified in inches.

We now have quite a few devices where we just round and hope, based on the assumption that it doesn't matter that our 6 inch device is actually 5.5 inches or whatever. We should switch the data files to use mm which have a better granularity and add an API to export those mm.

PR #896, opened 2025-08-02
Fix usable dimensions of the XP-PEN Artist 22R and 24 Pro

redstrate:
I didn't realize at the time that properties are supposed to be the usable area, *not* the physical dimensions of the device. Because #864 is not here yet, I rounded up for these.


redstrate:
> thx, merging. fwiw, doesn't matter since afaik nothing actually uses those values

Indeed, I realized that right after making the PR that libinput calculates it a different way!

whot:
> Indeed, I realized that right after making the PR that libinput calculates it a different way!

libinput gets it from the absinfo min/max and resolution (one reason why we need resolution for tablets btw). you can also get it from udev via ID_INPUT_WIDTH_MM and ID_INPUT_HEIGHT_MM but that property *may* be wrong if we have evdev overrides in place (rare enough for tablets at least)

PR #904, opened 2025-08-20
layouts: rearrange the Intuos Pro 3rd L/M buttons

whot:
The current layout is too far on the left/right edge, causing the right set of buttons to be nonconfigurable in GNOME's OSD, see e.g. [GNOME Settings#3500](https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/3500)

Re-do the layouts so the button configuration is below the dials. Ideally the GNOME OSD is smart enough so this isn't an issue to start with but since we have loads of empty space here anyway, let's make this a simple fix here until the rest can catch up.


whot:
I haven't done the S because that one looks workable (I think) and I'm lazy :)

PR #912, opened 2025-09-23
Added uniq identifier to Huion Kamvas Pro 13 (GT-133)

higginsdragon:
Later Huion Kamvas Pro 13 (GT-133) models have the productId 006d (of course) so I added the uniq identifier to the .tablet file. Tested working on my local system.

A word of note: For some reason `udevadm info` (v258) did not show a UNIQ attribtue, but `udevadm info -a` returned `ATTRS{uniq}=="HUION_M182_190116"`

The wiki might need to be updated to reflect this, not sure which version caused the change.


whot:
tha nks!

Issue #767, opened 2024-09-19
Support non-mangled Wacom tool IDs


jigpu:
The Linux kernel maintainers are open to the idea of eventually [removing tool ID mangling](https://lore.kernel.org/linux-input/20240903182633.870892-1-jason.gerecke@wacom.com/). Before we do this, we need to add the non-mangled versions of these IDs into libwacom so that applications can continue to work properly.

For reference, here is a complete list of all the affected pens / IDs. This list was made by finding all IDs in libwacom.stylus with the 4th nybble set to ‘0'. Versions of these IDs with the following nybbles right-shifted need to be added (e.g. mangled "0x10002" should have an un-mangled "0x1002" copy).
~~~
[0x10002] / [0x1000a]    # UD Pen (udpen)
[0x20802]                # Inking Pen
[0x40802] / [0x4080a]    # Classic Pen
[0x80842]                # Pro Pen 3D (mobilestudio)
[0x100802] / [0x10080a]  # Grip Pen (propengen2)
[0x100804] / [0x10080c]  # Art Pen (propengen2)
[0x100842] / [0x10084a]  # Pro Pen Slim (mobilestudio)
[0x100902] / [0x10090a]  # Airbrush Pen (propengen2)
[0x120802]               # Inking Pen (intuos5)
[0x140802] / [0x14080a]  # Classic Pen (propengen2)
[0x160802] / [0x16080a]  # Pro Pen (propengen2)
[0x180802] / [0x18080a]  # "Pen" (dth2242)
~~~

I can submit a patch that adds the non-mangled IDs easily enough, but I wonder if I should wait for #761 to be resolved first?

Also, is there a preference for having both mangled and non-mangled IDs exist in the same "libwacom.stylus" file, or would we want to move one of the two sets into a new stylus definition file?

whot:
#761 should be easy enough to implement, let's do this first please. I don't think we need a new `.stylus` file for 12 ids.

whot:
I don't think we have the actual aliases merged for those pens yet so while the infrastructure is now in place the actual fix required isn't.

PR #922, opened 2025-10-15
data: add unmangled stylus IDs to wacom.stylus

Pinglinux:
Years ago, an extra '0' accidently landed on the 4th nybble of the stylus IDs in Wacom kernel driver. We decided to fix the issue in the kernel (https://github.com/linuxwacom/libwacom/issues/767). But, we	want libwacom to be prepared for the change.

The "Mangled tool ID for kernel < 6.20"	comments in the	patch is only a placeholder. We will replace 6.20 with the actual kernel version that the mangled IDs are replaced by the correct ones.

### DIFF
--- a/data/wacom.stylus
+++ b/data/wacom.stylus
@@ -164,6 +164,10 @@ Axes=Tilt;Pressure;Distance;
 Type=Inking
 
 [0x56a:0x20802]
+# Mangled tool ID for kernel < 6.20
+AliasOf=0x56a:0x2802
+
+[0x56a:0x2802]
 # Intuos4, 5 and Cintiq 21UX2, 22HD, 24HD
 Name=Inking Pen
 Buttons=0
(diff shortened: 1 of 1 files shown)


data/wacom.stylus
```diff
@@ -298,63 +310,91 @@ Axes=Tilt;Pressure;Distance;RotationZ;
...
 [0x56a:0x100804]
+# Mangled tool ID for kernel < 6.20
+AliasOf=0x56a:0x10804
+
+[0x56a:0x10804]
 # Intuos4, 5 and Cintiq 21UX2, 22HD, 24HD
 Name=Art Pen
 Group=propengen2
-PairedStylusIds=0x56a:0x10080c;
+PairedStylusIds=0x56a:0x1080c;
```
whot: I think we'll need to keep the old `PairedStylusIds` for the now-aliased old pens. Otherwise your old-stylue 0x100804 pen ID tries to pair with a new-stylue 0x1080c pen which is wrong.
```
[0x56a:0x100804]
# Mangled tool ID for kernel < 6.20
AliasOf=0x56a:0x10804
PairedStylusIds=0x56a:0x10080c;
```

This should combine things correctly.

data/wacom.stylus
```diff
@@ -298,63 +310,91 @@ Axes=Tilt;Pressure;Distance;RotationZ;
 Type=Marker
 
 [0x56a:0x100804]
+# Mangled tool ID for kernel < 6.20
```
whot: is it going to be in 6.20? Right now we're in the 6.18 cycle, so there's a bit of prediction going on? :)

whot (changes requested):

PR #925, opened 2025-10-20
improved support for XP Pen Deco Pro v1 series

higginsdragon:
The XP Pen Deco Pro v1 series (M, MW, S, SW) all had very basic .tablet files, and no support for the buttons in udev-hid-bpf.

The button support will work with udev-hid-bpf pull request 208:
https://gitlab.freedesktop.org/libevdev/udev-hid-bpf/-/merge_requests/208

I updated the data/xp-pen-deco-pro-mw.tablet file and associated .svg file. The same SVG was used for the S and SW, so I made minor modifications to those so they could remain compatible and the tests could pass. In theory the udev-hid-bpf addition should work for all 4 tablets since the layouts are identical.

I don't have the descriptor names for the S and SW, I don't know if they would be "9 inch PenTablet" or "10 inch PenTablet" or possibly something else. Left a note at the top of the files indicating how support can be added in the future. I guess it would be possible to add both identifiers to the file but that seems messy.


data/xp-pen-deco-pro-mw.tablet
```diff
@@ -20,7 +22,7 @@
 [Device]
 Name=Hanvon Ugee Technology Co.,Ltd Deco Pro MW
 ModelName=
-DeviceMatch=usb|28bd|0934
+DeviceMatch=usb|28bd|0904|11 inch PenTablet;usb|28bd|0904|11 inch PenTablet Keyboard;usb|28bd|0904|11 inch PenTablet Mouse;usb|28bd|0934|11 inch PenTablet;usb|28bd|0934|11 inch PenTablet Keyboard;usb|28bd|0934|11 inch PenTablet Mouse
```
whot: are those usb IDs not unique? because if they are unique we don't need the name matches - those are only for devices where the pid is shared.

PR #931, opened 2025-10-29
Elan 42ab support

luyatshimbalanga:
This commit renames the incorrect labelled  wacom-isdv4-42ab.tablet to elan-42ab.tablet

### DIFF
--- a/data/elan-42ab.tablet
+++ b/data/elan-42ab.tablet
@@ -1,5 +1,4 @@
-# Dell Inspiron 7445 14 2-in-1
-# Sensor Type: AES
+# ELAN touchscreen/pen sensor present in the Dell Inspiron 7445 14 2-in-1
 # Features: Touch (Intergrated), Tilt
 #
 # sysinfo.BRJPJZKkp3.tar.gz
@@ -14,7 +13,7 @@ Class=ISDV4
 Width=12
 Height=7
 IntegratedIn=Display;System
-Styli=@isdv4-aes;
+Styli=@isdv4-aes;@generic;
 
 [Features]
 Stylus=true


data/elan-42ab.tablet
```diff
@@ -8,13 +7,12 @@
 
 [Device]
 Name=Dell Inc. Inspiron 14 7445 2-in-1
-ModelName=
-DeviceMatch=i2c|04f3|42ab
+ModelName=ELAN3233:00 04F3:42AB
+DeviceMatch=i2c|04f3|42ab;i2c|04f3|42ab Stylus
```
whot: this one looks wrong, the third entry is the PID but you're using a string here. And for devices that have a unique PID we don't need a name entry anyway, so you can skip this change.

data/elan-42ab.tablet
```diff
@@ -8,13 +7,12 @@
 
 [Device]
 Name=Dell Inc. Inspiron 14 7445 2-in-1
-ModelName=
-DeviceMatch=i2c|04f3|42ab
+ModelName=ELAN3233:00 04F3:42AB
```
whot: You can skip this line, the `ModelName` is for devices that have a marketing name and a model name, e.g. a Wacom Intuos Pro may be model number PTK-123. This doesn't really apply here since it's just the firmware nubmers here.

data/elan-42ab.tablet
```diff
@@ -8,13 +7,12 @@
...
+DeviceMatch=i2c|04f3|42ab;i2c|04f3|42ab Stylus
 Class=ISDV4
 Width=12
 Height=7
 IntegratedIn=Display;System
-Styli=@isdv4-aes;
```
whot: so it doesn't identify any stylus tools? If so it should use the `@generic` group

data/wacom-isdv4-42ab.tablet
```diff
@@ -8,13 +7,12 @@
...
 Name=Dell Inc. Inspiron 14 7445 2-in-1
-ModelName=
-DeviceMatch=i2c|04f3|42ab
+ModelName=ELAN3233:00 04F3:42AB
+DeviceMatch=i2c|04f3|42ab;i2c|04f3|42ab Stylus
 Class=ISDV4
 Width=12
 Height=7
 IntegratedIn=Display;System
-Styli=@isdv4-aes;
```
luyatshimbalanga: I found out @isdv4-aes only lists styli with one button when using Dell PN7522w which was 2 buttons and no physical eraser. I tried @generic-no-eraser but it seems missing tilt and pressure. @generic seems a logical choice.

luyatshimbalanga:
All set.  I found out the recent Dell Pro Premium Active Pen PN7522w is unlisted in wacom.stylus 

sudo python3 show-stylus.py
Using "ELAN3233:00 04F3:42AB Stylus": /dev/input/event7
Using stylus file(s): /usr/share/libwacom/lenovo.stylus, /usr/share/libwacom/wacom.stylus
Tool id 0x0 serial 0x1 in-proximity: False 
^CTerminating
Unknown stylus id 0x0. New entry needed
Styli=

I will open a separate issue about identifying that stylus.

luyatshimbalanga:
I am. The committed changes look good =)

PR #942, opened 2025-11-29
Add XP Pen Deco02 Support

HannahLP:
_Note: This is largely a duplicate of an historic pull request; I had made a mess of the repository, so pulled together a cleaner request from the master branch of my fork for use here._

This request adds support for the XP-Pen Deco02, based on my recent modifications made in the most recent [udev-hid-bpf release](https://gitlab.freedesktop.org/libevdev/udev-hid-bpf/-/releases).

This version is tested as working with the appropriate udev-hid-bpf program running, with GNOME's configuration tool correctly displaying and managing the inputs of the device.

The only present remaining issue is that this tablet supports exactly one pen - the P05, which supports only one button.
As far as I can tell, there is no generic one-button with eraser stylus option, and I unfortunately could not find where this stylus definition lives in the repository. As such, any configuration for the pen is presently going to be unclear.

### DIFF
--- a/data/layouts/xp-pen-deco02.svg
+++ b/data/layouts/xp-pen-deco02.svg
@@ -1,0 +1,165 @@
+<?xml version="1.0" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg
+   id="xp-pen-deco02"
+   width="10in"
+   height="5.6300001in"
+   style="color:#000000;stroke:#7f7f7f;fill:none;stroke-width:.25;font-size:8"
+   version="1.1"
+   viewBox="0 0 254 143.002"
+   xmlns="http://www.w3.org/2000/svg" >
+  <title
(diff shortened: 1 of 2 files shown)


data/layouts/xp-pen-deco02.svg
```diff
@@ -0,0 +1,165 @@
...
+       class="A Label"
+       x="66.5"
+       y="19"
+       style="text-anchor:start">A</text>
+  </g>
+  <g
+     id="DialGroup">
+      <circle
+     id="Dial"
+     class="Dial"
```
whot:
```suggestion
     class="TouchDial Dial"
```
I think this should do the trick.

whot:
>  unfortunately could not find where this stylus definition lives in the repository

`data/wacom.stylus` for wacom pens, etc. The stylis are assigned to groups and then the tablets reference the groups (some of which like the generic ones happen to only have one pen in it).

whot (changes requested):
The .tablet file LGTM, thanks, but the svg has at least one error, let's see what the CI says after adding the `TouchDial` class. Once those are fixed, happy to merge it.

whot (approved):

Issue #951, opened 2026-01-02
There is no generic stylus with tilt and distance but without eraser


ticotaco72:
- **Device name**: Huion Insipiroy Dial Q620M
- **Device model identifier**: Q620M
- **libwacom version**: git master

- [x ] I understand that libwacom does **not** affect whether the device works (see [Troubleshooting](https://github.com/linuxwacom/libwacom/wiki/Troubleshooting#my-tablet-doesnt-work))

# Bug description
I wanted to fill some info in `data/huion-inspiroy-q620m.tablet` but there is no generic stylus with pressure, distance and tilt but no eraser. How should that be added? would it be just:
```
[0x0:0xffffc]
Name=General Pen with no Eraser Plus
Group=generic-no-eraser-plus
Buttons=2
Axes=Tilt;Pressure;Distance;
Type=General
```

or similar?

whot:
It's basically just that but I think at this point we should introduce a `huion.stylus` file and populate that with the group of huion-supported generic styli and give them all to the same group. Pick some random-ish id that doesn't overlap with the other generic pens and done. Then switch the various huion data files to use those generic groups.

From our callers we won't be able to do much other than list the pens since (afaik) huion devices do not support identifying the pens. Maybe that's different with the new pens?

I also don't know the compatibility of the various new pens with older devices, do we need to group those separately?

whot:
Noteworthy: #984 adds a 3 buttons + eraser pen. This isn't what you need but it's roughly the best we can do. We *could* add a generic no-eraser pen but unless the device is limited to just that one pen and none other (and I'm pretty sure you could use other pens on the same device) then it'd just be limiting.

Unfortunately, until we can distinguish huion pens at runtime there isn't anything we can do.

whot:
I'm going to close this one as, effectively, wontfix. Until we can identify individual styli on Huion devices we simply can't fix this. #984 is the closest we get in a generic way but it's near the opposite of what you actually want/need.

You can work around this by putting a custom .stylus files in the lookup paths and using that but it's a cosmetic issue and barely worth the effort IMO.

PR #952, opened 2026-01-02
Update huion-inspiroy-q620m.tablet

ticotaco72:
Button I is inside Dial, but I believe there are no hardware modes. I am not sure if it should be described as in my patch. Dimensions taken from Huion website. I assumed there is only one version as to product id. Added previously submitted by someone sysinfo. Additionaly no exact stylus match is defined #951


whot (commented):
lgtm, minor change please, thanks

whot:
thanks!

PR #960, opened 2026-02-18
Add 2 more buttons to XP-Pen Artist 12

redstrate:
We emulate the touch bar as two additional buttons for this tablet.

See https://gitlab.freedesktop.org/libevdev/udev-hid-bpf/-/merge_requests/222


redstrate:
For some reason KDE Plasma is showing 3 buttons, this is probably our bug but I should look into it.

PR #962, opened 2026-02-23
data: Extend support for Lenovo Yoga X1 Gen8

hantangantan:
here a different variant for the Lenovo Yoga X1 Gen8 display


whot:
thanks!

PR #967, opened 2026-03-23
Switch tablet dimensions from inches to mm

### LINKED ISSUE #864: Switch Width/Height to mm

whot:
Change the Width/Height fields in our .tablet files to mm instead of inches and add a new API to get that value in mm, the existing API remains for backwards compatibility.

For easier transition print a warning if we have a width/height under 20 because that can't possibly be a mm value.

Closes #864


Pinglinux:
Patchset looks good. Thank you, Peter!

PR #978, opened 2026-05-16
Add support for Xencelabs Pen Display 16

BartHageman:


data/xencelabs-pen-display-16.tablet
```diff
@@ -0,0 +1,22 @@
+# Xencelabs Pen Display 16
+#
+# sysinfo.8KkSFPuMvY.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/618
+
+[Device]
+Name=hanvon ugee Xencelabs Pen Display 16
```
whot:
```suggestion
Name=Xencelabs Pen Display 16
```

This is going to be the name displayed to the users (despite whatever the firmware says) so I think we should go with the packaging name here. The `hanvon ugee` prefix is probably just there because they didn't clean up the firmware and/or for a user agent-like string matching in the driver.

whot (changes requested):
Almost ready, thanks. Please squash the requested change into the current commit, we cannot squash on merge here (iirc)

BartHageman:
Hope this works. One more detail that I forgot about:
The tablet has two different pens. One standard, one thin pen. Thin pen matches the generic pretty well since it only has two buttons and an eraser, but the regular one has three buttons. Would that require making a new .stylus file?

(forgot about it until now since I only use the thin one 😅)

whot:
>  The tablet has two different pens. One standard, one thin pen. Thin pen matches the generic pretty well since it only has two buttons and an eraser, but the regular one has three buttons. Would that require making a new .stylus file?

In theory yes but the main issue we have with this right now is that afaict we have nothing in userspace to be able to tell the pens apart when they come into proximity. We don't know the tool IDs and whether they're exposed by the tool at all so right now all of userspace just knows "pen in proximity".

What we should do is add a generic 3 button pen with eraser and switch all huion/ugee/* devices to that and let the user figure out the mismatch - better to have 3 buttons to configure when the pen only has 2 than preventing configuration of the third button. Never got around to that though.

whot (approved):

PR #981, opened 2026-05-21
Add support for Wacom HID 53fd Pen and Lenovo Yoga Pen

eliasvogel:
Added support for Wacom HID 53fd Pen with Data/wacom-hid-53fd-pen.tablet    
Added support for Lenovo Yoga Pen in Data/lenovo.stylus

Meson tests are all ok. They also caught an uppercase hex which I corrected.

After creating these, I found that  #949 uses the same tablet and pen, but is not the same laptop model as mine. I have decided to still make this pull request since the other has not been merged for almost half a year and I believe my PR adheres better to the project's conventions.

### DIFF
--- a/data/lenovo.stylus
+++ b/data/lenovo.stylus
@@ -24,3 +24,12 @@ Buttons=1
 EraserType=Button
 Axes=Tilt;Pressure
 Type=Mobile
+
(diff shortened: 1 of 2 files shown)


eliasvogel:
Commits are now squashed

PR #984, opened 2026-05-25
Add a generic 3 button + eraser pen

whot:
This sits on top of #983 

This covers the Huion PW600 device which is the most flexible device
that huion devices can support - it's a three-button pen with a tail-end
eraser. But so far we cannot know which pen is being used so we need to
implement this as a generic pen and hope that it doesn't confuse the
user too much.

The order (and fake tool ID) of pen and eraser shouldn't matter but we
give the pen a lower tool ID than the eraser. This means during typical
listing the pen comes first which helps implementations that don't
correctly check whether the tool is an eraser ~~(e.g. GNOME Settings
as of GNOME 50).~~ *edit: this is #985*

The tablets switch to the new pen are the ones listed on Huion's website [^1]: 
> Compatible with Kamvas Pro 19 and Kamvas Pro 27, Kamvas 16 (Gen 3) and Kamvas 13 (Gen 3)

This produces false positives if a user doesn't have that tool - we
still cannot identify Huion pens at runtime. But better to have
configuration for buttons that don't exist than lacking configurations
for buttons that do exist.

[^1]: https://store.huion.com/au/products/battery-free-pen-pw600

### DIFF
--- a/data/huion-kamvas-13-gs1333.tablet
+++ b/data/huion-kamvas-13-gs1333.tablet
@@ -42,7 +42,7 @@ Width=305
 Height=178
 Layout=huion-kamvas-13-gs1333.svg
 IntegratedIn=Display
-Styli=@generic-no-eraser;
+Styli=@generic-3btn-with-eraser;
 
 [Features]
 Stylus=true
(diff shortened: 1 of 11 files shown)


whot:
Probably noteworthy here: I did get tempted to make this a huion-specific pen (i.e. 256c vendor id) but the amount of work required here, in libinput, in the wayland protocol, in the compositor and the clients (e.g. GNOME Settings) for what ends up being zero benefit is simply not worth it.

PR #986, opened 2026-05-25
Export libwacom_stylus_lookup()

whot:
This is the vendor ID-enabled version of the deprecated libwacom_stylus_get_for_id().

libwacom_get_styli() is insufficient for GNOME Settings and libinput - both need to be able to find a stylus definition without having a WacomDevice as reference (in libinput that's an implementation detail that could probably be changed but GNOME Settings loads stylus definitions even when the stylus is not currently in proximity and thus truly has no device to be used as reference).

Assisted-by: Claude:claude-opus-4-6

### DIFF
--- a/libwacom/libwacom.c
+++ b/libwacom/libwacom.c
@@ -1754,6 +1754,19 @@ libwacom_stylus_get_for_stylus_id(const WacomDeviceDatabase *db,
 }
 
 LIBWACOM_EXPORT const WacomStylus *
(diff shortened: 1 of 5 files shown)


whot:
ftr, not 100% sure this is the right API, we have `libwacom_list_devices_from_database()` and for consistency we should probably have the same for styli instead

whot:
Closing this in favour of #987

PR #993, opened 2026-06-18
Add support for Lenovo Yoga 9 2-in-1 14IPH11

realRobotix:


data/lenovo.stylus
```diff
@@ -33,3 +33,12 @@ Buttons=1
 EraserType=Button
 Axes=Tilt;Pressure
 Type=Mobile
+
+[0x56a:0x544b]
```
whot, line 37: Just noticed this broke the CI but somehow managed to pass the merge queue anyway. Which made me realise: this does not look like the correct stylus id, it is the pid of the tablet but the styli themselves usually are 0x8*.

Can you please verify this? The `libwacom-show-stylus` tool (in the git repo) makes this easiest, otherwise it's the `ABS_MISC` value when you bring the tool into proximity, see e.g. `libinput record`.

realRobotix: libwacom-show-stylus: Tool id 0x8d15 serial 0x80201888

I unfortunately don't have the time to send a patch right now but i've tested it and libwacom-show-stylus recognizes it as isdv4-aes now with [0x56a:0x8d15]. Gnome picks it up too now, thanks!

PR #994, opened 2026-06-19
Add Huion Inspiroy H610Pro v2

kortschak:
Associated udev rules.

/etc/udev/rules.d/99-huion-h610pro-v2.rules 
```
# Huion Inspiroy H610Pro v2: prevent touch strip and dial from appearing as tablets
SUBSYSTEM=="input", ATTRS{idVendor}=="256c", ATTRS{idProduct}=="006d", ATTR{device/name}=="HUION Huion Tablet Touch Strip", ENV{ID_INPUT_TABLET}=""
SUBSYSTEM=="input", ATTRS{idVendor}=="256c", ATTRS{idProduct}=="006d", ATTR{device/name}=="HUION Huion Tablet Dial", ENV{ID_INPUT_TABLET}=""
```

### DIFF
--- a/data/huion-inspiroy-h610pro-v2.tablet
+++ b/data/huion-inspiroy-h610pro-v2.tablet
@@ -1,0 +1,27 @@
+# Huion
+# Inspiroy H610Pro v2
+#
+# sysinfo.fcHjk6pvOB.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/630
+
+[Device]
+Name=HUION Inspiroy H610Pro v2
+ModelName=Inspiroy H610Pro v2
+DeviceMatch=usb|256c|006d||HUION_T184
+Width=10
+Height=6
+Layout=huion-h610-pro.svg
+Styli=@generic-no-eraser;
+IntegratedIn=
+
+[Features]
+NumStrips=0
+NumRings=0
+Reversible=true
+Stylus=true
+Touch=false
+TouchSwitch=false
+
+[Buttons]
+Left=A;B;C;D;E;F;G;H
+EvdevCodes=BTN_0;BTN_1;BTN_2;BTN_3;BTN_4;BTN_5;BTN_6;BTN_7


data/huion-inspiroy-h610pro-v2.tablet
```diff
@@ -0,0 +1,27 @@
+# Huion
+# Inspiroy H610Pro v2
+#
+# sysinfo.fcHjk6pvOB.tar.gz
+# https://github.com/linuxwacom/wacom-hid-descriptors/issues/630
+
+[Device]
+Name=HUION Inspiroy H610Pro v2
+ModelName=Inspiroy H610Pro v2
+DeviceMatch=usb|256c|006d||HUION_T184
```
whot, line 10: looks like we already have this device from #965 but the match unintentionally included the firmware date suffix `_190313`. If you remove that from `./data/huion-h610-pro-v2.tablet` does it work for you? the rest is the same afaict.

If it works, can you update the PR (or file a new one) removing that suffix please? Thanks

kortschak: Is matching performed with prefix matching, or exact matching. If I remove the suffix and it's exact matching the currently supported tablet firmware version will cease to work. Also, for the follow-up I'd like to fix the file name to follow conventions; is that too much diff noise?

whot: It should ignore the suffix, see [the code here](https://github.com/linuxwacom/libwacom/blob/d9624fbfd84d0191a27a9bef8268cf300d2b598c/libwacom/libwacom.c#L272).
Maybe a stray underscore?

Oh, and note that #965 also included the name in the match (`HUION Huion Tablet`), if your device has a different name (e.g. a `Pen` or `Pad` suffix) that'll be the reason for the failed match. Remove the name match too so it matches your `DeviceMatch` line and that should work, assuming no other bug has snuck in somewhere.

> is that too much diff noise?

Just keep that change as a separate commit and you can make all the noise you want ;)

PR #1002, opened 2026-07-20
Added Samsung Galaxy Book3 360

LunarEclipse363:
See https://github.com/linuxwacom/wacom-hid-descriptors/issues/633

### DIFF
--- a/data/sambook3-360.tablet
+++ b/data/sambook3-360.tablet
@@ -1,0 +1,20 @@
+# Samsung
+# Galaxy Book3 360 NP730QFG
+#
+# UD-EMR Stylus Compatible
+#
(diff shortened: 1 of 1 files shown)


whot (approved):
If you remove the `wip` from the commit message this one seems ready :)

LunarEclipse363:
I think it should be ready, I wanted to double check everything again but it's probably not happening any time soon :sweat_smile: 

I'm not 100% sure if the stylus group is correct, this panel seems to support both styli with an eraser and without one, I would guess that the correct option is the with eraser in such case.

LunarEclipse363:
FWIW I only amended the commit title to remove the [WIP]

PR #1003, opened 2026-07-21
Add support for Lenovo Yoga 9 2-in-1 14IPH11

realRobotix:
Fixes #999.


whot:
thanks for the quick fix!

PR #1009, opened 2026-07-31
Add support for XP-Pen Deco SE

YangtseSu:
Add support for XP-Pen Deco SE


YangtseSu:
Deco SE and Deco mini7W V2 share the same device ID. Their parameters are also largely consistent, but the SE is only released in China.

whot:
thanks!

